PR #56, opened 2026-02-11
Focus terminal when clicking a worktree in the sidebar

barrettj:
## Summary
- Clicking a worktree in the sidebar now immediately transfers focus to the terminal surface
- Previously, focus stayed on the sidebar's NSTableView after clicking, causing the first arrow key to navigate the sidebar instead of going to the terminal
- Detects mouse clicks via `NSApp.currentEvent` and sets the pending terminal focus flag, reusing the existing `pendingTerminalFocusWorktreeIDs` mechanism
- Keyboard navigation in the sidebar is unaffected (arrow keys still navigate the list as before)

## Test plan
- [ ] Click a worktree in the sidebar — terminal should have focus immediately (arrow keys go to terminal, not sidebar)
- [ ] Use keyboard arrows to navigate sidebar — should still navigate worktrees without stealing focus to terminal
- [ ] Create a new worktree — terminal should still auto-focus as before
- [ ] Existing tests pass (`make test`)


barrettj:
The issue is the opposite - I clicked a workspace expecting to then be able to interact with my terminal (it had choices, I wanted the second so hit down arrow) and it instead navigates the list (but only once or twice then down arrow throws me into one of the terminals) Sent from my iPhoneOn Feb 11, 2026, at 10:24 AM, khoi ***@***.***> wrote:﻿khoi left a comment (supabitapp/supacode#56)
can't replicate, I can navigate up and down fine using the keyboard

khoi:
please use the hotkey or the command pallete to quickly switch and type command, I thikn mouse click and up and down arrows to navigate is working as intended

PR #75, opened 2026-02-17
fix(terminal): sync color scheme with Ghostty theme preference

slvssb:
## Summary

- Respects user's Ghostty theme preference when syncing color scheme
- Correctly parses partial dual theme formats (e.g., `light:tokyonight,dark:catppuccin`)
- Extracts GhosttyConfigUpdater into a dedicated utility for cleaner architecture
- Passes initial color scheme to GhosttyRuntime before surface creation

## Test plan

- [x] Verify app appearance setting syncs to Ghostty config
- [x] Test with Ghostty `theme` set to a single theme name
- [x] Test with Ghostty `theme` set to dual theme format (`light:X,dark:Y`)
- [x] Test with Ghostty `theme` set to partial dual format (only light or only dark)
- [x] Verify terminal surfaces use correct theme on creation

### DIFF
--- a/supacode/Infrastructure/Ghostty/GhosttyConfigUpdater.swift
+++ b/supacode/Infrastructure/Ghostty/GhosttyConfigUpdater.swift
@@ -11,6 +11,8 @@ enum GhosttyConfigUpdater {
   private enum ParsedTheme {
     case none
     case single(String)
+    case partialLight(String)  // Only light:Theme specified
+    case partialDark(String)  // Only dark:Theme specified
     case dual(light: String, dark: String)
   }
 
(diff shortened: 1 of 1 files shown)


slvssb:
Hi, this adds support for a dual theme configuration in Ghostty when only one theme is specified, or when no theme is specified at all. If both themes are already specified, it simply respects the existing setup and doesn't override anything.

I'm not entirely sure whether this affects all users, but I personally spent quite a bit of time earlier figuring out that the theme had to be explicitly set up in Ghostty. This change helps avoid that confusion for new users, especially those who haven't installed or configured Ghostty before but rely on system appearance settings via Supacode.

I thought this would be helpful for that reason, but of course, I'll respectfully leave the final call to you.

PR #159, opened 2026-03-19
Parallelize worktree loading and prioritize last-focused repo on startup

onevcat:
## Problem

App startup is blocked by serial `wt ls --json` calls in `loadRepositoriesData`. Each call spawns a login shell (`/bin/zsh -l -c ...`), which takes ~370ms due to shell startup (loading `.zshrc`, mise, plugins, etc.). With N repositories this becomes ~370ms × N — **4.8 seconds for 13 repos** in my case, stopping me from using the app smoothly.

## Benchmark (13 repos, Apple Silicon)

| Phase | Duration |
|-------|----------|
| `init()` total | 66ms |
| Persistence load | 6ms |
| **Each repo `wt ls --json`** | **~360-390ms** |
| **`loadRepositoriesData` total (serial)** | **4.79s** |

The `init()` and persistence loading are fast. The bottleneck is entirely in the serial worktree fetching loop.

### 1. Parallel loading with `TaskGroup`

Replace the serial `for root in roots` loop with `withTaskGroup` so all repos fetch concurrently. Results are sorted by input index after collection to preserve deterministic ordering — this matters for fallback selection logic (`firstAvailableWorktreeID`).

### 2. Priority loading of last-focused repo

When `lastFocusedWorktreeID` is set, emit an early `repositoriesLoaded` as soon as that repo's fetch completes. This lets the UI appear immediately with the selected worktree while remaining repos load in the background.

Priority matching uses fetched worktree IDs (not path prefix), so it works correctly even when linked worktrees live outside the repository root.

## Results

| Metric | Before | After |
|--------|--------|-------|
| **UI usable** | 4.83s | **0.39s** |
| All repos loaded | 4.83s | ~1.27s |
| **Speedup** | baseline | **12x** |

## Test plan

- [x] All existing tests pass
- [x] New tests for priority loading with linked worktrees
- [x] New test verifying root order preservation when loads finish out-of-order
- [x] New test verifying fallback selection respects root order after repo removal
- [x] Lint passes

### DIFF
--- a/supacode/Features/Repositories/Reducer/RepositoriesFeature.swift
+++ b/supacode/Features/Repositories/Reducer/RepositoriesFeature.swift
@@ -259,6 +259,12 @@ struct RepositoriesFeature {
     let didPruneArchivedWorktreeIDs: Bool
   }
 
+  private struct RepositoryLoadBatch {
+    let repositories: [Repository]
+    let failures: [LoadFailure]
+    let didLoadPriorityRepository: Bool
+  }
+
   enum StatusToast: Equatable {
     case inProgress(String)
     case success(String)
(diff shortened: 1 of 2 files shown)


onevcat:
Maybe I guess a full cache would be better. I will open another PR soon.

PR #162, opened 2026-03-20
Add repository snapshot startup cache

onevcat:
## Summary

- Restore repositories from a standalone snapshot cache before the live refresh completes, allowing the UI to render immediately on launch
- Persist the snapshot only after a full successful repository load, keeping failed or partial loads from overwriting good cache data
- Add focused tests for snapshot payload validation, startup cache hit/miss behavior, and snapshot persistence

This is the final piece of a three-part startup performance series, building on #160 (parallel loading) and #161 (direct bundled `wt` execution).

### Before

https://github.com/user-attachments/assets/b9426f1f-553b-4109-958a-28a8c162fdf5

### After

https://github.com/user-attachments/assets/574cb141-1c1a-4018-8b46-3db41b2b771e

## Problem

App startup is blocked by worktree discovery — even after parallelizing and removing login-shell overhead, the app still needs to run `wt ls --json` for every repository before the UI can render. With 13 repositories, users still see a loading state for over a second.

## Approach

Introduce a lightweight JSON snapshot file (`~/.supacode/repository-snapshot.json`) that captures just enough data for the sidebar to render:

- Repositories in UI order (root path, display name)
- Worktree entries (name, detail string, working-directory path, `createdAt`)

The snapshot is loaded synchronously at startup and used to populate the repository list immediately. A normal live refresh always follows to pick up any changes since the last snapshot. The snapshot is only overwritten after a complete successful live load — partial or failed loads never corrupt it.

**What is NOT cached:** PR state, line changes, watcher state, notifications, `lastFocusedRepositoryID`. Selection restoration continues to use the existing `lastFocusedWorktreeID` persistence.

**Cache invalidation:** The snapshot is treated as a miss (and discarded) when:
- File is missing, empty, or has a schema version mismatch
- JSON decode fails
- Any cached repository root path or worktree path no longer exists on disk

## Benchmark (13 repos, Apple Silicon)

| Optimization | Time to UI usable |
|---|---|
| Baseline (serial, login shell) | 6.28s |
| + Parallel loading (#160) | 4.45s |
| + Direct bundled `wt` (#161) | 1.20s |
| **+ Snapshot cache (this PR)** | **0.41s** |
| **Total speedup** | **~15x** |

The remaining ~400ms is dominated by other startup work (window restoration, terminal initialization, etc.) and is already fast enough for a smooth launch experience.

## Design decisions

- **Standalone file, not in settings.json**: Cache decode failures stay isolated from user settings. The file can be deleted safely with no migration burden.
- **No TTL**: The cache is only a startup accelerator. Freshness comes from the unconditional live refresh that always runs after cache restore.
- **Path-existence validation**: Guards against stale entries from removed repos or worktrees without needing a version scheme tied to repo changes.

## Test plan

- [x] All existing tests pass
- [x] New tests for snapshot encode/decode round-trip
- [x] New tests for cache-hit startup flow (UI renders before live refresh)
- [x] New tests for cache-miss startup flow (falls back to normal loading)
- [x] New tests for snapshot persistence after successful live load
- [x] Lint passes

### DIFF
--- a/supacode/Clients/Repositories/RepositoryPersistenceClient.swift
+++ b/supacode/Clients/Repositories/RepositoryPersistenceClient.swift
@@ -15,6 +15,8 @@ struct RepositoryPersistenceClient {
   var saveWorktreeOrderByRepository: @Sendable ([Repository.ID: [Worktree.ID]]) async -> Void
   var loadLastFocusedWorktreeID: @Sendable () async -> Worktree.ID?
   var saveLastFocusedWorktreeID: @Sendable (Worktree.ID?) async -> Void
+  var loadRepositorySnapshot: @Sendable () async -> [Repository]?
+  var saveRepositorySnapshot: @Sendable ([Repository]) async -> Void
 }
 
 extension RepositoryPersistenceClient: DependencyKey {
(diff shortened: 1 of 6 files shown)


khoi:
i dont know whether introducing cache for this is worth it

khoi:
yes, let's trade correctness for a bit of slower speed. ur other 2 prs already address speed a bit, keep it coming, ty

PR #179, opened 2026-03-27
fix cyrillic layout switch key leak

yaroslavyaroslav:
## Summary

Fixes a keyboard input bug where switching to a Cyrillic/Russian layout could leak raw keyboard protocol fragments like `[32;1:3u` into the Codex terminal input.

The root cause was that `keyDown` was already suppressed when the macOS input source changed, but the matching `keyUp` and modifier release events were still being forwarded to Ghostty. This change adds a narrow suppression window for those follow-up events and covers it with tests.

## Verification

- `xcodebuild test -project supacode.xcodeproj -scheme supacode -destination "platform=macOS" -only-testing:supacodeTests/ GhosttySurfaceViewTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY="" -skipMacroValidation -quiet`
- `make build-app`
- `make test`
- `make format`

### DIFF
--- a/supacode/Infrastructure/Ghostty/GhosttySurfaceView.swift
+++ b/supacode/Infrastructure/Ghostty/GhosttySurfaceView.swift
@@ -11,6 +11,48 @@ final class GhosttySurfaceView: NSView, Identifiable {
     let length: UInt64
   }
 
+  struct KeyboardLayoutChangeSuppression: Equatable {
+    static let relevantModifiers: NSEvent.ModifierFlags = [.shift, .control, .option, .command]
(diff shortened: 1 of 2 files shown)


khoi:
i think the fix should stay smaller here. [32;1:3u looks like the space keyup, so i would only suppress the matching keyup after the layout switch and leave flagsChanged alone. swallowing modifier changes feels risky and it dont match upstream ghostty, which already moved away from that because it broke bindings.

PR #211, opened 2026-04-02
Fix pane focus sync without terminal auto-focus

yaroslavyaroslav:
So the issue was that on repo/worktree switch follow mouse to focus were still having issues sometimes (i.e. it were falling into undefined state with few panes being focused).

## Summary
- prevent panes from becoming Ghostty-focused unless they are also the real AppKit first responder
- keep the focused pane id as intent and reconcile when focus callbacks fire
- add coverage for the no-auto-focus policy and the responder-driven pane switch flow

## Verification
- xcodebuild test -quiet -project supacode.xcodeproj -scheme supacode -destination "platform=macOS" -only-testing:supacodeTests/TerminalRenderingPolicyTests -only-testing:supacodeTests/WorktreeTerminalManagerTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY="" -skipMacroValidation
- make build-app


sbertix:
Hey 👋 could you get a bit more into details about the issue?

yaroslavyaroslav:
Yep, focus still had buggy behavior when switching to the new worktree/repo by mouse. 

This way first responder still stays the section panel yet some terminal pane already became active, so when mouse were moved to different pane directly — it happened to be active as well yet the initially active never were marked as non active. 

So you have to click on few panes to resolve that collision. 

I think it could be quite stable reproduced in to have three vertical splits and move mouse to the bottom one after clicking that repo from a repo list. 

Often happens to me when I select  a repo down the list and then moving my pointer to through the bottom panel up to the top automatically (I mean unintentionally)

UPD: this is the core idea of PR some struct refactoring happened after linter failed on too many attributes on a function.

sbertix (commented):
Nice fix 💪 just a few comments

sbertix:
Also, I'm not sure this fixes it entirely when switching quickly between two worktrees: I've been able to replicate the issue (two terminals active side by side). Could you double check, please? 🙇‍♂️

sbertix:
This last commit now works with the fast switching, but has the "opposite" issue: we now get some worktrees that get a focused terminal on switch, and some that don't at all  until you actively tap/type

yaroslavyaroslav:
I'm about to finish this bit one way or another, sorry for it taking that long from my side guys. It's just a little bit a lot on my plate since recently. 

So thank you for drafting it rather closing. 

So my last step was to compare this pane management logic with the original one ghosty has. Still need to review it carefully and maybe migrate there entirely after all.

PR #225, opened 2026-04-08
fix Ghostty background opacity toggle action handling

### LINKED ISSUE #180: Toggle Background Opacity is not working
Hello, I was trying to use `Toggle Background Opacity` action from the Command Palette - similar to how I use it in Ghostty, but it's not doing anything.

Image

vadimi:
Handle Ghostty toggle-background-opacity action in the surface bridge and add a per-window opaque override in GhosttySurfaceView. This lets users temporarily force an opaque terminal background when transparency is enabled, while preserving fullscreen behavior.

This should fix https://github.com/supabitapp/supacode/issues/180

### DIFF
--- a/supacode/Infrastructure/Ghostty/GhosttySurfaceBridge.swift
+++ b/supacode/Infrastructure/Ghostty/GhosttySurfaceBridge.swift
@@ -68,6 +68,9 @@ final class GhosttySurfaceBridge {
       return onMoveTab?(action.action.move_tab) ?? false
     case GHOSTTY_ACTION_TOGGLE_COMMAND_PALETTE:
       return onCommandPaletteToggle?() ?? false
+    case GHOSTTY_ACTION_TOGGLE_BACKGROUND_OPACITY:
+      surfaceView?.toggleBackgroundOpacity()
+      return true
     case GHOSTTY_ACTION_GOTO_WINDOW,
       GHOSTTY_ACTION_TOGGLE_QUICK_TERMINAL,
       GHOSTTY_ACTION_CLOSE_ALL_WINDOWS:
--- a/supacode/Infrastructure/Ghostty/GhosttySurfaceBridge.swift
+++ b/supacode/Infrastructure/Ghostty/GhosttySurfaceBridge.swift
@@ -70,7 +70,7 @@ final class GhosttySurfaceBridge {
       return onCommandPaletteToggle?() ?? false
     case GHOSTTY_ACTION_TOGGLE_BACKGROUND_OPACITY:
       surfaceView?.toggleBackgroundOpacity()
-      return true
+      return surfaceView != nil
     case GHOSTTY_ACTION_GOTO_WINDOW,
       GHOSTTY_ACTION_TOGGLE_QUICK_TERMINAL,
       GHOSTTY_ACTION_CLOSE_ALL_WINDOWS:
--- a/supacode/Infrastructure/Ghostty/GhosttySurfaceBridge.swift
+++ b/supacode/Infrastructure/Ghostty/GhosttySurfaceBridge.swift
@@ -69,8 +69,7 @@ final class GhosttySurfaceBridge {
     case GHOSTTY_ACTION_TOGGLE_COMMAND_PALETTE:
       return onCommandPaletteToggle?() ?? false
     case GHOSTTY_ACTION_TOGGLE_BACKGROUND_OPACITY:
-      surfaceView?.toggleBackgroundOpacity()
-      return surfaceView != nil
+      return surfaceView?.toggleBackgroundOpacity() ?? false
     case GHOSTTY_ACTION_GOTO_WINDOW,
       GHOSTTY_ACTION_TOGGLE_QUICK_TERMINAL,
       GHOSTTY_ACTION_CLOSE_ALL_WINDOWS:
--- a/supacode/Infrastructure/Ghostty/GhosttySurfaceView.swift
+++ b/supacode/Infrastructure/Ghostty/GhosttySurfaceView.swift
@@ -92,6 +92,7 @@ final class GhosttySurfaceView: NSView, Identifiable {
   private var lastSurfaceFocus: Bool?
   private var eventMonitor: Any?
   private var notificationObservers: [NSObjectProtocol] = []
+  private var isBackgroundOpaque = false
   private var prevPressureStage: Int = 0
   private lazy var cachedScreenContents = CachedValue<String>(duration: .milliseconds(500)) {
     [weak self] in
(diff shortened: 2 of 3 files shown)


supacode/Infrastructure/Ghostty/GhosttySurfaceBridge.swift
```diff
@@ -68,6 +68,9 @@ final class GhosttySurfaceBridge {
       return onMoveTab?(action.action.move_tab) ?? false
     case GHOSTTY_ACTION_TOGGLE_COMMAND_PALETTE:
       return onCommandPaletteToggle?() ?? false
+    case GHOSTTY_ACTION_TOGGLE_BACKGROUND_OPACITY:
+      surfaceView?.toggleBackgroundOpacity()
+      return true
```
sbertix: `isBackgroundOpaque` is per-view, but `applyWindowBackgroundAppearance()` modifies window-level properties. With splits, another surface's call to `applyWindowBackgroundAppearance()` would revert the toggle silently cause its own `isBackgroundOpaque` is still `false`. Could we move this state to window-level instead? Wdyt?

sbertix: NIT, other actions return the closure result (or `false` when nil), but this returns `true` unconditionally even if `surfaceView` is nil. Could we do `return surfaceView != nil`? (But this improvement is negligible)

supacode/Infrastructure/Ghostty/GhosttySurfaceView.swift
```diff
@@ -92,6 +92,7 @@ final class GhosttySurfaceView: NSView, Identifiable {
   private var lastSurfaceFocus: Bool?
   private var eventMonitor: Any?
   private var notificationObservers: [NSObjectProtocol] = []
+  private var isBackgroundOpaque = false
```
sbertix: Are we considering resetting `isBackgroundOpaque` when `ghosttyRuntimeConfigDidChange` fires? Cause if the user toggles opaque then changes the config, the flag persists and they'd need to toggle twice 🤔

supacode/Infrastructure/Ghostty/GhosttySurfaceView.swift
```diff
@@ -416,6 +417,13 @@ final class GhosttySurfaceView: NSView, Identifiable {
     window.backgroundColor = runtime.backgroundColor().withAlphaComponent(1)
   }
 
+  func toggleBackgroundOpacity() {
+    guard runtime.backgroundOpacity() < 1 else { return }
+    guard let window, !window.styleMask.contains(.fullScreen) else { return }
```
sbertix: NIT, both guards are re-checked inside `applyWindowBackgroundAppearance()`. Could we add a short comment explaining they prevent toggling `isBackgroundOpaque` when it'd have no visible effect? Otherwise it reads like duplication at first glance.

sbertix (commented):
Thanks for the contribution 🙇‍♂️

supacode/Infrastructure/Ghostty/GhosttySurfaceBridge.swift
vadimi: ah yes, that totally makes sense. I've checked ghostty and it does it per tab, but I like per window much better. I've added `isBackgroundOpaque` on the NSWindow level. I haven't found any other custom state properties in the repo, so please let me know if you prefer another approach. Ghostty does it on the windowController level, but it's not used in this repo.

supacode/Infrastructure/Ghostty/GhosttySurfaceView.swift
vadimi: I've checked ghostty and it looks like it doesn't do it when I reload the config.

supacode/Infrastructure/Ghostty/GhosttySurfaceView.swift
vadimi: done

supacode/Infrastructure/Ghostty/GhosttySurfaceBridge.swift
```diff
@@ -68,6 +68,9 @@ final class GhosttySurfaceBridge {
       return onMoveTab?(action.action.move_tab) ?? false
     case GHOSTTY_ACTION_TOGGLE_COMMAND_PALETTE:
       return onCommandPaletteToggle?() ?? false
+    case GHOSTTY_ACTION_TOGGLE_BACKGROUND_OPACITY:
+      surfaceView?.toggleBackgroundOpacity()
+      return surfaceView != nil
```
sbertix: My bad, missing it earlier. 🙇‍♂️ Could we make `toggleBackgroundOpacity` return `Bool` and handle it like everything else?

supacode/Infrastructure/Ghostty/GhosttySurfaceView.swift
sbertix: Could we add some `surfaceLogger.warning(...)` for the no-ops and `surfaceLogger.debug(...)` for the early returns?

supacode/Infrastructure/Ghostty/GhosttySurfaceView.swift
```diff
@@ -416,6 +417,15 @@ final class GhosttySurfaceView: NSView, Identifiable {
     window.backgroundColor = runtime.backgroundColor().withAlphaComponent(1)
   }
 
+  func toggleBackgroundOpacity() {
+    // we keep these preconditions here so we do not mutate per-window opacity state 
+    // when toggling would have no visible effect.
```
sbertix:
```swift
// Guard against a missing window, and skip toggling when it would
// have no visible effect (fully opaque config or fullscreen mode).
```

Possibly reads better, but 100% NIT.

supacode/Infrastructure/Ghostty/GhosttySurfaceView.swift
```diff
@@ -1440,6 +1450,26 @@ final class GhosttySurfaceView: NSView, Identifiable {
...
+      objc_setAssociatedObject(
+        self,
+        &GhosttySurfaceWindowBackgroundOpaqueKey.value,
+        newValue,
+        .OBJC_ASSOCIATION_RETAIN_NONATOMIC
+      )
+    }
+  }
+}
+
```
sbertix, line 1469: It can work like this, and I should have been more specific with what I meant, but I feel like this is over-the-top for what we need, where we can get window-level behavior by simply having `var isBackgroundOpaque` on `GhosttyRuntime`, since it's a single shared instance. Wdyt?

vadimi: absolutely, done

supacode/Infrastructure/Ghostty/GhosttySurfaceView.swift
vadimi: sure, updated the comment

supacode/Infrastructure/Ghostty/GhosttySurfaceBridge.swift
vadimi: done!

supacode/Infrastructure/Ghostty/GhosttySurfaceView.swift
vadimi: done, added logs

sbertix:
Thank you 🙇‍♂️

vadimi:
thanks for your guidance @sbertix!

PR #227, opened 2026-04-10
Add CLI tool with socket transport and related skills

sbertix:
## Summary

- **supacode-cli**: ArgumentParser-based CLI that communicates with the running app via Unix domain socket. Supports worktree, tab, surface, repo, and settings commands.
- **Socket protocol**: Command (deeplink dispatch with response FD) and query (read-only state snapshots) message types with full FD lifecycle management.
- **Settings**: CLI install/uninstall (symlink to `/usr/local/bin`), skill install for Claude Code and Codex, `AutomatedActionPolicy` replacing the old boolean toggle.
- **CLI/Deeplink reference windows**: In-app reference views for both CLI commands and deeplink URLs, accessible from Developer settings.
- **Renamed**: `DeeplinkCheatsheet` → `DeeplinkReference` throughout.

## Test plan

- [ ] Install CLI from Developer settings, run `supacode worktree list` from a terminal
- [ ] Run `supacode tab new -i "echo hello"` — should prompt for confirmation (unless policy is `.always`)
- [ ] Run `supacode surface split -d v` from inside a Supacode terminal
- [ ] Kill and relaunch Supacode, verify CLI commands from old terminals still work (stale socket fallback)
- [ ] Pass `--id` with a UUID that already exists — should get an error, not a silent collision
- [ ] Open CLI Reference and Deeplink Reference from Developer settings


khoi (approved):
lets' go

PR #228, opened 2026-04-11
Add mcp server for agent orchestration

BitWonka:
# Summary

Adds an MCP layer so a manager agent can spawn and coordinate coding agents across worktrees/repos. Builds on the existing hook infrastructure for real-time status tracking. The idea came from wanting to delegate work across multiple repos where the initial agent has almost all the context on a feature since that was the initial point of discussion and also making small changes in other related repos like docs. 

Note: This PR might has overlap with #227

- Add an MCP server that lets a manager agent orchestrate coding agents (supagents) across repositories/worktrees,  spawning, messaging, reading screens, and monitoring status in real-time
- Standalone supacode-mcp binary (stdio MCP transport) communicates with Supacode via persistent Unix domain socket
- 6 MCP tools: list_worktrees, get_worktree_status, spawn_supagent, send_message, read_screen, list_notifications
- SurfaceID-scoped completion tracking with debounced resolution, rolling inactivity timeout, and proper continuation cleanup
- Hook events pushed to connected clients in real-time via server.log()
- Settings UI for server toggle + per-agent MCP registration
- 20 new tests: 9 SocketClient completion tracking, 11 MCPServerInstaller

# Possible Future Enhancements
- Dedicated Orchestrator panel
- Create worktree
- Skills for direction
- Run agents without permissions

# Test plan

- Enable MCP in Settings > Coding Agents, verify socket starts
- Start Claude Code/Codex session, verify list_worktrees returns repos
- spawn_supagent in a worktree: tab opens, response returned with surface_id
- send_message with wait=true: agent response returned
- read_screen: terminal content returned
- Concurrent agents in same worktree tracked independently
- Disable MCP: socket stops, binary enters degraded mode on next Claude session

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,6 +65,9 @@ fastlane/test_output
 Frameworks/GhosttyKit.xcframework
 Resources/ghostty
 Resources/terminfo
+Resources/supacode-mcp
+supacode-mcp/.build/
+supacode-mcp/Package.resolved
 *.profraw
 .env
 .DS_Store
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -2,9 +2,11 @@ strict: true
 
 included:
   - supacode
+  - supacode-mcp
   - supacodeTests
 excluded:
   - ThirdParty/ghostty
+  - supacode-mcp/.build
 
 disabled_rules:
   - file_length
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ else
   $(error Unknown FORMAT "$(FORMAT)". Use xcsift, xcpretty, or none)
 endif
 .DEFAULT_GOAL := help
-.PHONY: build-ghostty-xcframework build-app run-app install-dev-build archive export-archive format lint check test bump-version bump-and-release log-stream
+.PHONY: build-ghostty-xcframework build-mcp build-app run-app install-dev-build archive export-archive format lint check test bump-version bump-and-release log-stream
 
 help:  # Display this help.
 	@-+echo "Run make with one of the following targets:"
@@ -53,7 +53,11 @@ $(GHOSTTY_BUILD_OUTPUTS):
 	mkdir -p "$$terminfo_dst"; \
 	rsync -a --delete "$$terminfo_src/" "$$terminfo_dst/"
 
-build-app: build-ghostty-xcframework # Build the macOS app (Debug)
+build-mcp: # Build the MCP orchestrator binary
+	cd supacode-mcp && swift build -c release
+	cp -f supacode-mcp/.build/release/supacode-mcp Resources/supacode-mcp
+
+build-app: build-ghostty-xcframework build-mcp # Build the macOS app (Debug)
 	bash -o pipefail -c 'xcodebuild -project supacode.xcodeproj -scheme supacode -configuration Debug build -skipMacroValidation -clonedSourcePackagesDirPath "$(SPM_CACHE_DIR)" 2>&1 $(FORMATTER)'
 
 run-app: build-app # Build then launch (Debug) with log streaming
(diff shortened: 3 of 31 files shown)


khoi:
@BitWonka we're about to add skills to the mix, any reason why u would prefer an MCP server instead?

BitWonka:
@khoi by skills i think you referring to the CLI? 

well the MCP can give us opportunity for more complex behaviour since it can be a bi-directional and persistent connection. For example, 

from worktree B: "spawn supagent in worktree X and wait for response" it will start a new claude/codex session in worktree X and wait for the response. which gets delivered by the mcp. 

then you can ask it to have a discussion with the supagent in worktree X and they will keep on chatting away amongst each other. the mcp server latches on to the hooks we already have for notifications so they can be sent to the agent that invoked the other agent. it paves the way for autonomous orchestration. 

I havent gone through the CLI PR so i dont fully unerstand the current scope but the CLI is generally more so for single operations from agentic perspective or interactive mode for human operations such as claude or codex. for agentic use, the MCP is clear winner, for human use the CLI would be the obvious choice so they could easily coexist, just would need to align for consistency. I wasnt aware the CLI was in the works until I was about to PR and mad respect for @sbertix, i did reach out to him to give a heads up and to coordinate.

sbertix:
I think the orchestration stuff (spawn, block until done, screen read, event streaming) is definitely a good addition, but I still feel like a more agnostic approach (which Supacode kept till now) possibly makes more sense, without limiting the improvements to "agents" alone. 

That said, I think there's value in trying to integrate these on top of the socket layer in #227. There we already have PID-scoped discovery, stale pruning, permissions, FD lifecycle, and everything apart from an actual `--wait` instruction (which is key to this PR). 

I'm a bit more hesitant on the MCP binary tbh. It couples to the app without being bundled with it (unlike the CLI), so we'd need version checks, protocol sync across targets: the protocol file is already duplicated with a "keep in sync" comment. Not sure it's worth it when an MCP server could just wrap the CLI/skills (skills which are already shipped with #227) externally. Wdyt?

My suggestions overall would be:

| | Component | Action | Notes |
|---|---|---|---|
| ✅ | `spawn_supagent` | **Already in #227** | `supacode tab new --input` / `supacode surface split --input` and this is not agent-limited so it can work with whatever command |
| ✅ | `send_message` | **Already in #227** | `supacode surface focus --input "..."` and not agent-limited either |
| 🔨 | Completion tracking | **Add `--wait` flag** | On `tab new`, `surface split`, `surface focus` and all `--input` accepting subcommands in general |
| 🔨 | `read_screen` | **Add subcommand** | `supacode surface read` (or whatever) |
| ⏸️ | `list_notifications` | **Add subcommand** | `supacode worktree notifications` but I would wait for this, cause notifications might benefit from a slight rework |
| 🔨 | Event streaming | **Add to socket layer** | Persistent connection mode for `--wait` support |
| 🗑️ | `MCPSocketServer` | **Drop** | Reuse #227's socket layer |
| 🗑️ | `MCPSocketProtocol` (duplicated) | **Drop** | No second target to sync |
| 🗑️ | `supacode-mcp` binary | **Drop** | Can be built externally wrapping the CLI if needed |
| 🗑️ | `MCPServerInstaller` | **Drop** | No MCP binary to register |
| ✅ | Permission gating | **Already in #227** | `AutomatedActionPolicy` covers new commands |
| ✅ | Reconnection handling | **Already in #227** | PID-scoped discovery + auto-launch |

BitWonka:
I think this is a fair direction and makes alot of sense. I’m less attached to the separate MCP binary if the same capabilities can live cleanly on top of the socket layer. I will close this PR and port over the missing functionalities in another branch utilizing the CLI and socket layer.

PR #232, opened 2026-04-13
Fix off-by-one terminal column count with right-aligned PS1

### LINKED ISSUE #231: Error on right aligned PS1
Hi Supacode,

I found a little UI issue with Supacode:

First the visual proof:

Supacode:
Image

Ghostty/iTerm/Kitty/...:
Image

See that the closing `]`, the last character in the line, drops to the next line.  It feels like an off by one error.

I'm using starship.rs for my shell prompt, and this is the format:

```
format = "$directory$git_branch${custom.git_branch_description}$git_commit$git_state$git_metrics$git_status$fill$nix_shell$aws$line_break$jobs$battery$time$status$os$container$shell$character"
```

(notice that `$aws` comes right before `$line_break`)

pepegar:
## Summary

- Use `scrollView.contentSize` instead of `scrollView.bounds` for PTY size reporting, matching upstream Ghostty's `synchronizeCoreSurface()` pattern
- Add `safeAreaInsets` override returning zero on `GhosttySurfaceScrollView`, matching upstream
- Delegate surface size change notifications through the scroll wrapper when available, ensuring the correct content-area width is always used

When `contentSize` differs from `bounds` (due to content insets or safe area adjustments), the PTY was told a width slightly too large, causing the last character on a full-width prompt line to wrap.

## Test plan

- [ ] Open a worktree terminal with a starship prompt using `$fill` (right-aligned content that fills the full terminal width)
- [ ] Verify the closing `]` or last character stays on the same line instead of wrapping
- [ ] Resize the window and confirm columns update correctly
- [ ] Verify scrollback works as before
- [ ] Test with both overlay and legacy scroller styles (System Settings > Appearance > Show scroll bars)

Closes #231

### DIFF
--- a/supacode/Infrastructure/Ghostty/GhosttySurfaceView.swift
+++ b/supacode/Infrastructure/Ghostty/GhosttySurfaceView.swift
@@ -357,7 +357,7 @@ final class GhosttySurfaceView: NSView, Identifiable {
     }
     updateScreenObservers()
     updateContentScale()
-    updateSurfaceSize()
+    notifySizeChanged()
     applyWindowBackgroundAppearance()
   }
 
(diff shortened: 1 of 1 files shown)


pepegar:
sorry folks, this was autocreated by the agent, don't review

PR #233, opened 2026-04-13
Fix right-aligned PS1 width reporting

### LINKED ISSUE #231: Error on right aligned PS1

khoi:
- route Ghostty surface size changes through the scroll wrapper when present
- report `scrollView.contentSize.width` to the PTY instead of `scrollView.bounds.width`

Closes #231

### DIFF
--- a/supacode/Infrastructure/Ghostty/GhosttySurfaceView.swift
+++ b/supacode/Infrastructure/Ghostty/GhosttySurfaceView.swift
@@ -357,7 +357,7 @@ final class GhosttySurfaceView: NSView, Identifiable {
     }
     updateScreenObservers()
     updateContentScale()
-    updateSurfaceSize()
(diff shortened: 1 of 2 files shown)


sbertix (approved):
💪

PR #241, opened 2026-04-14
Respect `split-preserve-zoom` config when cycling splits

Whoaa512:
Supacode reimplements split management in `WorktreeTerminalState` rather than using Ghostty's `BaseTerminalController`. This means the upstream `split-preserve-zoom` config option (added in Ghostty v1.3.0) was never read — cycling splits while zoomed always unzoomed.

This PR adds support for `split-preserve-zoom = navigation` so that `goto_split:next`/`goto_split:previous` transfers the zoom to the target pane instead of unzooming, matching upstream Ghostty's behavior in `BaseTerminalController`.

### Changes

- **`GhosttyRuntime.swift`** — read `split-preserve-zoom` from ghostty config (bit 0 = navigation)
- **`WorktreeTerminalState.swift`** — `gotoSplit` checks config: preserves zoom when enabled, unzooms when not
- **`SplitTreeTests.swift`** — test covering zoom transfer on focus cycling

### Reference

Upstream implementation: [`BaseTerminalController.swift` L645-L651](https://github.com/ghostty-org/ghostty/blob/main/macos/Sources/Features/Terminal/BaseTerminalController.swift#L645-L651)

### DIFF
--- a/supacode/Features/Terminal/Models/WorktreeTerminalState.swift
+++ b/supacode/Features/Terminal/Models/WorktreeTerminalState.swift
@@ -21,6 +21,7 @@ final class WorktreeTerminalState {
 
   let tabManager: TerminalTabManager
   private let runtime: GhosttyRuntime
+  @ObservationIgnored private let splitPreserveZoomOnNavigation: () -> Bool
   private let worktree: Worktree
   @ObservationIgnored
   @SharedReader private var repositorySettings: RepositorySettings
@@ -63,8 +64,14 @@ final class WorktreeTerminalState {
   var onCommandPaletteToggle: (() -> Void)?
   var onSetupScriptConsumed: (() -> Void)?
 
-  init(runtime: GhosttyRuntime, worktree: Worktree, runSetupScript: Bool = false) {
+  init(
+    runtime: GhosttyRuntime,
+    worktree: Worktree,
+    runSetupScript: Bool = false,
+    splitPreserveZoomOnNavigation: (() -> Bool)? = nil
+  ) {
     self.runtime = runtime
+    self.splitPreserveZoomOnNavigation = splitPreserveZoomOnNavigation ?? { runtime.splitPreserveZoomOnNavigation() }
     self.worktree = worktree
     self.pendingSetupScript = runSetupScript
     self.tabManager = TerminalTabManager()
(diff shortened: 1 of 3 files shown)


supacode/Features/Terminal/Models/WorktreeTerminalState.swift
```diff
@@ -574,8 +574,13 @@ final class WorktreeTerminalState {
...
       if tree.zoomed != nil {
-        tree = tree.settingZoomed(nil)
-        trees[tabId] = tree
+        if runtime.splitPreserveZoomOnNavigation() {
+          let nextNode = tree.root?.node(view: nextSurface)
+          tree = tree.settingZoomed(nextNode)
+        } else {
+          tree = tree.settingZoomed(nil)
+        }
+        updateTree(tree, for: tabId)
```
sbertix: We're testing the tree-level zoom transfer in `SplitTreeTests` (💪), but imho we're missing coverage for the actual branching logic in `performSplitAction(.gotoSplit)`, i.e., the `splitPreserveZoomOnNavigation()` check. Could we add a test that exercises both paths (preserve vs clear)? Cause right now if someone accidentally flips the condition, nothing would catch it.

supacode/Infrastructure/Ghostty/GhosttyRuntime.swift
```diff
@@ -557,6 +557,14 @@ final class GhosttyRuntime {
     return true
   }
 
+  func splitPreserveZoomOnNavigation() -> Bool {
+    guard let config else { return false }
+    var value: CUnsignedInt = 0
+    let key = "split-preserve-zoom"
+    guard ghostty_config_get(config, &value, key, UInt(key.count)) else { return false }
+    return value & (1 << 0) != 0
```
sbertix, line 568: This bitmask check `(value & (1 << 0) != 0)` is relying on Ghostty's internal encoding for "on navigation". Could we add a short comment here explaining the bit layout? I'm worried 2 weeks from now this is gonna look pretty cryptic without context. Wdyt?

Whoaa512: yup makes total sense. can add comment with permalink to ghostty source

Whoaa512:
Addressed review feedback in 1710303:

- added performSplitAction(.gotoSplit) coverage for preserve vs clear zoom config paths
- added bitset note + Ghostty source permalink for split-preserve-zoom

Validation: scoped swiftlint passed. Local xcodebuild/make build-app currently fail in swift-composable-architecture under my local Xcode/toolchain so relying on CI here

PR #245, opened 2026-04-15
Add Kiro CLI as third coding agent with hooks and CLI skill

b0x42:
Adds [Kiro CLI](https://kiro.dev) as a third coding agent alongside Claude Code and Codex in Settings → Coding Agents.

image

### New files
- **`KiroHookSettings.swift`** — Hook entry struct (`command` + `timeout_ms`) and payloads for progress (`userPromptSubmit` → busy on, `stop` → busy off) and notification (`stop` → notify) events
- **`KiroHookSettingsFileInstaller.swift`** — File installer for Kiro's flat hook format: install, uninstall, and containsMatchingHooks
- **`KiroSettingsInstaller.swift`** — Targets `~/.kiro/agents/kiro_default.json`, creates default agent config when file is absent
- **`KiroSettingsClient.swift`** — TCA dependency client wrapping the installer

### Modified files
- **`AgentHooksInstallState.swift`** — Added `.kiroProgress`, `.kiroNotifications` to `AgentHookSlot`
- **`SkillAgent.swift`** — Added `.kiro` case with `configDirectoryName: ".kiro"`
- **`SettingsFeature.swift`** — Wired Kiro state, dependency, check/install/uninstall effects
- **`DeveloperSettingsView.swift`** — Added Kiro section with progress, notifications, and CLI skill rows
- **`CLISkillContent.swift`** — Added `kiroSkillMd` (SKILL.md with YAML frontmatter)
- **`CLISkillInstaller.swift`** — Added `.kiro` case
- **`AgentHookSocketServer.swift`** — Added `assistant_response` decoding (see below)

### Assets
- **`kiro-mark`** image set (SVG from kiro.dev)

### Why `kiro_default.json`
Kiro has no global hooks file — hooks live per-agent in `~/.kiro/agents/<name>.json`. The `kiro_default` agent is what runs in Supacode terminals, so that's the only file we target. The built-in `kiro_default` has no file on disk; creating `kiro_default.json` overrides it entirely, so the installer writes a complete config (name, tools, resources, hooks) — not just hooks.

### Why a separate `KiroHookSettingsFileInstaller`
Kiro's hook format is flat:
```json
{ "hooks": { "stop": [{ "command": "...", "timeout_ms": 10000 }] } }
```
Claude/Codex use a grouped format:
```json
{ "hooks": { "Stop": [{ "hooks": [{ "type": "command", "command": "...", "timeout": 10 }] }] } }
```
Rather than generalizing `AgentHookSettingsFileInstaller` and risking regressions in Claude/Codex, this isolates the format difference in a separate installer.

### Socket server: `assistant_response` field
Kiro sends the agent's final response in `assistant_response` rather than `message` (Claude) or `last_assistant_message` (Codex). Extended the notification body fallback chain:
```swift
let body = payload.message ?? payload.lastAssistantMessage ?? payload.assistantResponse
```
Without this, Kiro notifications would have empty bodies.

### Uninstall safety
Uninstall is a no-op if `kiro_default.json` doesn't exist, preventing creation of a broken config file that would override Kiro's built-in agent with missing fields.

## Verification
- `make check` passes (swift-format + swiftlint)
- `make test` passes (716 tests, 57 suites)


supacode/Clients/CodingAgents/KiroSettingsClient.swift
```diff
@@ -0,0 +1,44 @@
+import ComposableArchitecture
```
sbertix: This needs to move to `SupacodeSettingsShared/Clients/CodingAgents/` and become `public` (struct, init, properties, DependencyKey conformance), otherwise the `SupacodeSettingsFeature` module can't see it and we don't even build. The other two clients (`ClaudeSettingsClient`, `CodexSettingsClient`) already live there.

Also, when you move it, the `public init(...)` block has 4-space indentation while the rest of the project uses 2-space. Would fail `make format`.

SupacodeSettingsFeature/Reducer/SettingsFeature.swift
```diff
@@ -41,6 +41,9 @@ public struct SettingsFeature {
     public var claudeNotificationsState = AgentHooksInstallState.checking
     public var codexProgressState = AgentHooksInstallState.checking
     public var codexNotificationsState = AgentHooksInstallState.checking
```
sbertix, line 43: Are we skipping tests on this one? Cause the reducer now has 3 new state fields, 6 new switch cases, and the `.task` action checks Kiro hooks, but nothing in the test suite covers any of it. `taskChecksAllFourHookSlotsOnStartup` still asserts 4 slots (not 6), `receiveStartupHookChecks` doesn't assert `kiroProgressState` / `kiroNotificationsState` / `kiroSkillState`, and there's no dedicated test for `.agentHookInstallTapped(.kiroProgress)` and friends. Same for `KiroHookSettingsFileInstaller` and `KiroSettingsInstaller`, both have zero unit tests while their Claude/Codex counterparts have full suites. We should probably match that coverage, imho.

SupacodeSettingsShared/BusinessLogic/KiroSettingsInstaller.swift
```diff
@@ -0,0 +1,144 @@
...
+    guard fileManager.fileExists(atPath: settingsURL.path) else { return }
+    try fileInstaller.uninstall(
+      settingsURL: settingsURL,
+      hookEntriesByEvent: try KiroHookSettings.notificationHookEntriesByEvent()
+    )
+  }
+
+  // MARK: - Default agent config.
+
+  /// Creates `kiro_default.json` with the known built-in defaults when the file does not exist.
```
sbertix, line 101: Could we add a note on which Kiro version these defaults are based on? Cause this is the only installer that creates the entire agent config from scratch (Claude/Codex just merge hooks into existing files), and the comment already says "overrides the built-in entirely." If Kiro ships new defaults down the line, we'd want to know what we're comparing against. Something like:

```suggestion
  /// Creates `kiro_default.json` with the known built-in defaults when the file does not exist.
  /// The config mirrors Kiro's built-in `kiro_default` agent as of Kiro 1.x (2026-04). `tools: ["*"]`
  /// matches Kiro's default tool access. Creating this file overrides the built-in entirely, so we must
  /// include the full config (not just hooks).
```

Wdyt?

supacode/Features/Settings/Views/DeveloperSettingsView.swift
```diff
@@ -89,6 +89,42 @@ struct DeveloperSettingsView: View {
       } footer: {
         Text("Applied to `~/.codex`.")
       }
+      Section {
```
sbertix: 🤔 Are we sure we want Kiro at the same visual weight as Claude Code and Codex? Cause it's a newer integration and having it always-expanded might give the impression it's equally supported. Could we use `Section(isExpanded:)` with a default of `false` so it's collapsed by default? We'd lose the section footer ("Applied to `~/.kiro`."), but I think that's a fair trade for a cleaner hierarchy. How does it feel?

SupacodeSettingsShared/BusinessLogic/KiroHookSettings.swift
```diff
@@ -0,0 +1,86 @@
...
+  let hooks: [String: [KiroHookEntry]] = [
+    "stop": [
+      KiroHookEntry(command: KiroHookSettings.notify, timeoutMs: 10_000)
+    ],
+  ]
+}
+
+// MARK: - Payload support for flat format.
+
+nonisolated enum KiroHookPayloadSupport {
```
sbertix: `KiroHookPayloadSupport.extractHookEntries` is functionally identical to `AgentHookPayloadSupport.extractHookGroups`, same decode-to-JSONValue, same `objectValue["hooks"]` extraction, same iterate-and-collect. We could just reuse the existing one and delete `KiroHookPayloadSupport` entirely (~20 lines). Not blocking, but it's a quick win. I'm good with a follow-up too, tbh.

sbertix (changes requested):
Thanks 💪 just a few changes required, though.

supacode/Infrastructure/AgentHookSocketServer.swift
```diff
@@ -429,12 +429,14 @@ private nonisolated struct AgentHookPayload: Decodable {
   let title: String?
   let message: String?
   let lastAssistantMessage: String?
+  let assistantResponse: String?
```
sbertix: With `assistant_response` being the only field Kiro emits on `stop`, are we considering adding a decode test for it? Cause `parsesNotificationWithLastAssistantMessageFallback` covers the Claude/Codex branches but nothing exercises this new fallback, so if the field ever drifts the bell would silently come through with an empty body and we'd never notice in tests.

Probably also worth a precedence test (`message ?? lastAssistantMessage ?? assistantResponse`, all three present → `message` wins) just to lock the chain.

supacode/Infrastructure/AgentHookSocketServer.swift
```diff
@@ -386,7 +386,7 @@ final class AgentHookSocketServer {
       return nil
     }
 
-    let body = payload.message ?? payload.lastAssistantMessage
+    let body = payload.message ?? payload.lastAssistantMessage ?? payload.assistantResponse
```
sbertix: When all three body fields are nil, we end up delivering a notification with an empty body (consumer reduces it to `""`). If Kiro renames the field again we wouldn't see anything in `make log-stream` either. Could we log a warning + maybe synthesize a placeholder so we at least notice?

Also a one-liner above the fallback would help future-me a lot:
```swift
// Kiro emits assistant_response on stop; Claude uses message, Codex uses last_assistant_message.
```

SupacodeSettingsShared/BusinessLogic/KiroSettingsInstaller.swift
```diff
@@ -0,0 +1,144 @@
...
+  /// The config mirrors Kiro's built-in `kiro_default` agent as of Kiro 1.x (2026-04). `tools: ["*"]`
+  /// matches Kiro's default tool access. Creating this file overrides the built-in entirely, so we must
+  /// include the full config (not just hooks).
+  private func ensureDefaultAgentConfig() throws {
+    guard !fileManager.fileExists(atPath: settingsURL.path) else { return }
+    try fileManager.createDirectory(
+      at: settingsURL.deletingLastPathComponent(),
+      withIntermediateDirectories: true
+    )
+    let defaultConfig: [String: JSONValue] = [
```
sbertix, line 112: The doc comment is super clear that we're overriding Kiro's built-in entirely and need the *full* config, but `KiroSettingsInstallerTests` only checks `fileExists`. If `name`, `tools: ["*"]`, the four `resources`, or `useLegacyMcpJson` ever drift from what we wrote, users silently lose Kiro functionality and the test suite stays green. Could we decode the written JSON and assert each of those fields? Otherwise the "complete config" invariant isn't really enforced anywhere.

SupacodeSettingsShared/BusinessLogic/KiroSettingsInstaller.swift
```diff
@@ -0,0 +1,144 @@
...
+      "name": .string("kiro_default"),
+      "tools": .array([.string("*")]),
+      "resources": .array([
+        .string("file://AGENTS.md"),
```
sbertix, line 116: Quick sanity check on `skill://.kiro/skills/**/SKILL.md`: `CLISkillInstaller` writes the Kiro skill to `~/.kiro/skills/<name>/SKILL.md` (absolute home), but Kiro's `skill://` scheme is typically project-relative iiuc. Have we verified this resolves to the home-installed skill? Cause if it doesn't, the SKILL.md row in DeveloperSettings will show "installed" but Kiro won't actually see it 🤔

supacodeTests/KiroHookSettingsFileInstallerTests.swift
```diff
@@ -0,0 +1,140 @@
...
+          "command": .string(AgentHookSettingsCommand.busyCommand(active: false)),
+          "timeout_ms": 10_000,
+        ]),
+      ],
+    ]
+  }
+
+  // MARK: - Install.
+
+  @Test func installIntoEmptyFileCreatesCorrectStructure() throws {
```
sbertix, line 42: We have nice coverage for round-trip and idempotency 💪, but I couldn't find the equivalent of `AgentHookSettingsFileInstallerTests.uninstallRemovesOnlyMatchingCommands`. Could we add one for the flat format? Cause if `isManaged` ever has a regression (or the legacy-command matcher false-positives), we'd silently delete a user's hand-written `hooks.stop` entry and the user would have no idea where it went.

SupacodeSettingsShared/BusinessLogic/KiroHookSettingsFileInstaller.swift
```diff
@@ -0,0 +1,185 @@
...
+
+  func install(
+    settingsURL: URL,
+    hookEntriesByEvent: @autoclosure () throws -> [String: [JSONValue]]
+  ) throws {
+    let settingsObject = try loadSettingsObject(at: settingsURL)
+    let hookEntries = try hookEntriesByEvent()
+    let commandsToPrune = Self.commands(from: hookEntries)
+    var mergedObject = settingsObject
+    var hooksObject = (mergedObject["hooks"]?.objectValue) ?? [:]
```
sbertix: Small policy divergence I noticed: when `hooks` is a non-object value, this installer falls back to `[:]` silently, while `AgentHookSettingsFileInstaller` throws `invalidHooksObject`. The `Errors.invalidHooksObject` closure here is also dead code as a result. We probably want both installers on the same policy, no? Either both lenient or both strict. Personally I'd lean strict so we don't silently bulldoze a malformed (but recoverable) file, but I'm good with either tbh.

SupacodeSettingsShared/BusinessLogic/KiroHookSettings.swift
```diff
@@ -0,0 +1,62 @@
...
+    case timeoutMs = "timeout_ms"
+  }
+}
+
+// MARK: - Progress hooks.
+
+// Kiro uses camelCase event names ("userPromptSubmit", "stop") unlike
+// Claude/Codex which use PascalCase ("UserPromptSubmit", "Stop").
+private nonisolated struct KiroProgressPayload: Encodable {
+  let hooks: [String: [KiroHookEntry]] = [
```
sbertix, line 56: Claude registers `UserPromptSubmit`, `Stop`, `PostToolUseFailure`, `SessionEnd` so the busy flag clears even on crashes/disconnects. Kiro only has `userPromptSubmit` (on) and `stop` (off). Does Kiro expose anything analogous for failure / session-end? Cause if it does, missing it means the sidebar gets stuck on busy when Kiro crashes. If it genuinely doesn't, a one-line comment saying so would save someone the same investigation later.

supacodeTests/KiroSettingsInstallerTests.swift
```diff
@@ -0,0 +1,119 @@
...
+    // Second install should not recreate the base config.
+    try installer.installProgressHooks()
+    let secondWrite = try Data(contentsOf: settingsURL)
+
+    // Content must still be valid JSON after second install.
+    #expect(try JSONDecoder().decode(JSONValue.self, from: secondWrite).objectValue != nil)
+    // Hooks section should still exist.
+    let json = try JSONDecoder().decode(JSONValue.self, from: secondWrite)
+    #expect(json.objectValue?["hooks"] != nil)
+    _ = firstWrite  // used
```
sbertix:
```swift
_ = firstWrite  // used
```
100% nitpicking, but the test name says "does not overwrite existing config" and `firstWrite` is never actually compared to `secondWrite` (and the `// used` comment is technically false). Could we either drop the variable or replace it with `XCTAssertEqual(firstWrite, secondWrite)` so the assertion matches the name? Otherwise a regression here would slip through.

sbertix (commented):
Nice work 💪 A few testing gaps + small things inline 🙇‍♂️

b0x42:
## Feedback addressed (2 commits)

**`skill://` URI fix (bug):** The `skill://` scheme in `kiro_default.json` was using project-relative paths (`skill://.kiro/skills/**/SKILL.md`) but `CLISkillInstaller` writes skills to `~/.kiro/skills/` (home directory). Per [Kiro docs](https://kiro.dev/docs/cli/skills/), custom agents need `skill://~/...` for global skills. Fixed to `skill://~/.kiro/skills/**/SKILL.md` and `skill://~/.kiro/steering/**/*.md`. Without this, the SKILL.md row showed "installed" but Kiro never actually loaded it.

**`assistant_response` decode tests:** Added `parsesNotificationWithAssistantResponseFallback` (Kiro-only field) and `messageFieldTakesPrecedenceOverAllFallbacks` (all three fields present → `message` wins) to lock the fallback chain.

**Nil-body warning log:** Added `SupaLogger.warning` when all body fields (`message`, `lastAssistantMessage`, `assistantResponse`) are nil, plus a comment explaining which agent uses which field. If Kiro renames the field again, it now surfaces in `make log-stream`.

**Full config assertions in `KiroSettingsInstallerTests`:** Tests now decode the written JSON and assert `name`, `tools: ["*"]`, `useLegacyMcpJson`, all 4 `resources` entries, and `hooks` — instead of just `fileExists`. Also fixed the `firstWrite`/`secondWrite` comparison (`#expect(firstWrite == secondWrite)` instead of `_ = firstWrite // used`).

**`uninstallRemovesOnlyMatchingCommands` for flat format:** Mirrors the existing Claude/Codex grouped-format test — installs managed hooks, manually adds a user hook (`echo user-hook`), uninstalls, asserts only the user hook remains.

sbertix (approved):
Thank you 💪

PR #257, opened 2026-04-19
Add folder (non-git) repository support + harden removal pipeline

### LINKED ISSUE #252: Feature Request: Possibility to use outside of a git repo
Spinning up claude inside of folders that are not tracked by git is something that i do often and not having the possibility to open a project on the side that is not tracked by git for me is a missed feature not an advantage.

sbertix:
## Summary

- Adds non-git **folder** repositories: open any directory through the repo picker and it renders as its own sidebar section with a single synthetic main worktree. Terminal tabs, hotkeys, scripts, command palette, and deeplinks all reuse the existing `SidebarSelection` / `WorktreeTerminalManager` plumbing. Git-only actions (archive, pin, unpin, new worktree) surface a folder-incompatible alert instead of silently no-op-ing.
- Refactors the sidebar UI into shared `SidebarItem*` views (`WorktreeRow` → `SidebarItemView`, `WorktreeRowsView` → `SidebarItemsView`) so git and folder rows share rendering.
- Rationalizes the delete pipeline around a typed batch aggregator (`ActiveRemovalBatch`, `RepositoryRemovalRecord`, `DeleteDisposition`). Bulk deletes now fire one terminal `.repositoriesRemoved` per batch, coalesce parallel trash failures into a single alert, and can't clobber concurrent git-section and folder batches.
- Splits `RemovalOutcome.failure` into `.failureSilent` (caller owns the alert — blocking-script, cancel, kind-flip) and `.failureWithMessage(String)` (aggregator owns the alert — `trashItem` errors). Makes the alert-clobber scenario unrepresentable: a script-failure alert sharing a batch with a trash failure is preserved; the trash errors are logged via `repositoriesLogger.warning` instead of overwriting the UI.
- Prunes `removingRepositoryIDs` + `activeRemovalBatches` in `applyRepositories` symmetric with the seven other reload-intersected trackers.
- Surfaces missing-on-disk folder roots through the same `LoadFailure` → `SidebarFailedRepositoryRow` pipeline git repos already use, via a new `gitClient.rootDirectoryExists` dependency. A folder deleted from Finder while Supacode was running now shows the error row instead of an empty folder.
- Logs ambiguous damaged-bare-clone paths (lastComponent ends in `.git` but HEAD/objects/refs trio is incomplete) in `Repository.isGitRepository(at:)` so the otherwise-silent misclassification is observable in telemetry without widening the classifier and creating false positives.
- Dedups the `folderIncompatibleAlert` copy across the reducer helper and the `AppFeature` deeplink handler via `FolderIncompatibleAction.alertCopy`.
- Moves the `seedRemovalBatch` test helper out of `#if DEBUG` in the main module and into the test target where `@testable import supacode` already exposes the internal state fields.
- Adds `.help()` to the Folder Settings context-menu button to match CLAUDE.md's tooltip mandate.
- Prunes `settingsFile.repositories` entries when a repository is removed so the script / run-config / open-action dicts don't leak forever.

## Test plan

- [x] `make check` (swift-format + swiftlint, strict mode)
- [x] `make test` — 849 tests pass, 12 known issues unchanged from the branch baseline
- [x] Open a non-git folder via the repo picker; confirm it renders a section with one selectable row
- [x] Hit Archive / Pin / Unpin on a folder row via hotkey and via deeplink — both show the folder-incompatible alert with identical copy
- [x] Bulk-delete two folders where one script fails and the other's trash throws — script-failure alert stays visible; trash failure is logged
- [x] Delete a folder from Finder while Supacode is running — refresh the sidebar — folder row shows the failed-repository error row with a "may have been moved or deleted" message
- [x] Rapid double-tap Delete on a folder with an empty delete script — second tap is a no-op, no duplicate analytics
- [x] Trigger a reload while a folder trash is in-flight — batch still drains; state stays consistent

Closes #252


bjufre:
After this update Supacode stopped detecting my bare repo as such and just shows a folder, and after the restart I lost the layout and panes I was working and stuck back to Zellij and old plain Ghostty 🙏 😢

sbertix:
> After this update Supacode stopped detecting my bare repo as such and just shows a folder

Sorry to hear that. Could you give me some more info on the structure of the repo so I can add some more regression testing? @bjufre 

> and after the restart I lost the layout and panes

When repos are rendered as folders they don't lose their metadata, they just hide all info which is no longer relevant, so in theory if you haven't made any other changes it should be back once it's fixed.

sbertix:
I've started a new release extending the name convention checks we used before, to rely more on the folder structure. `0.8.3` is being released rn. If you can try it to see if it fixes the issue, it would be greatly appreciated 🙇‍♂️ @bjufre

PR #269, opened 2026-04-23
Add tab renaming

bitomule:
## What

Enable renaming of tabs.

You can double tap the label or use the context menu option. I didn't understand the script tabs blocking so I tried not to interfere with any of that. Name is persisted so future sessions keep the custom name.

## Disclaimer
I read you prefer issues than vibe coded PRs. I used claude code on this one but manually supervised plan, code and tested manually. I think issue is small enough and a PR is helpfull in this case. Otherwise please let me know.

## How it works

**Model layer**
- `TerminalTabItem` gains `customTitle: String?` and a computed `displayTitle: String` (`customTitle ?? title`). `isTitleLocked` is left exclusively for blocking-script tabs — no semantic change.
- `TerminalTabManager.updateTitle()` gains a `customTitle == nil` guard so Ghostty surface-title callbacks don't overwrite a user rename.
- `TerminalTabManager.setCustomTitle(_:title:)` sets/clears the override without touching `isTitleLocked`.

**Persistence**
- `TerminalLayoutSnapshot.TabSnapshot` gains `customTitle: String?`. The field is `Optional` so existing `layouts.json` files without the key decode to `nil` — zero migration cost.
- `captureLayoutSnapshot()` serialises `customTitle` (nil for blocking-script tabs, same as the existing icon/tintColor normalisation).
- `restoreFromSnapshot()` calls `setCustomTitle` after recreating the tab if the snapshot carries a custom name.

**UI**
- `TerminalTabLabelView` renders `tab.displayTitle` instead of `tab.title`.
- `TerminalTabView` adds a `@FocusState`-managed `TextField` overlay that appears on double-tap (`TapGesture(count: 2)`). Enter commits, Escape cancels, focus loss commits. The label and close button are hidden (`opacity: 0 / allowsHitTesting: false`) while the field is active so there's no visual bleed-through.
- The rename action is threaded from `WorktreeTerminalTabsView` → `TerminalTabBarView` → `TerminalTabsView` → `TerminalTabsRowView` following the same closure-chain pattern as `closeTab` / `closeOthers`.
- `TerminalTabContextMenu` shows "Rename Tab" for non-locked tabs, wired to `editingTabId` state in `TerminalTabsRowView`.

**Tests**
- 4 new `TerminalTabManagerTests`: custom title overrides display title, does not set `isTitleLocked`, blocks Ghostty updates, clears correctly.
- 2 new `TerminalLayoutSnapshotTests`: `customTitle` round-trips through JSON encoding; missing key in legacy JSON decodes as `nil`.

## Smoke test checklist
- [x] Double-click a tab → TextField appears pre-filled with current title
- [x] Type new name + Enter → tab shows custom name
- [x] Run `vim` → custom name stays, Ghostty title ignored
- [x] Escape during edit → reverts, no change applied
- [x] Submit empty string → no change
- [ ] Double-click a blocking-script tab (Archive/Delete) → no edit field
- [x] Right-click → "Rename Tab" appears for user tabs, absent for script tabs
- [x] Quit + reopen → custom names survive across sessions

### DIFF
--- a/supacode/Features/Terminal/Models/TerminalTabManager.swift
+++ b/supacode/Features/Terminal/Models/TerminalTabManager.swift
@@ -49,7 +49,13 @@ final class TerminalTabManager {
   func updateTitle(_ id: TerminalTabID, title: String) {
     guard let index = tabs.firstIndex(where: { $0.id == id }) else { return }
     guard !tabs[index].isTitleLocked else { return }
+    guard tabs[index].customTitle == nil else { return }
     tabs[index].title = title
+  }
+
+  func setCustomTitle(_ id: TerminalTabID, title: String?) {
+    guard let index = tabs.firstIndex(where: { $0.id == id }) else { return }
+    tabs[index].customTitle = title
   }
 
   func unlockAndUpdateTitle(_ id: TerminalTabID, title: String) {
--- a/supacode/Features/Terminal/Models/TerminalTabManager.swift
+++ b/supacode/Features/Terminal/Models/TerminalTabManager.swift
@@ -7,6 +7,7 @@ import SupacodeSettingsShared
 final class TerminalTabManager {
   var tabs: [TerminalTabItem] = []
   var selectedTabId: TerminalTabID?
+  var pendingRenameTabID: TerminalTabID?
 
   private static let logger = SupaLogger("TabManager")
 
@@ -82,6 +83,7 @@ final class TerminalTabManager {
   func closeTab(_ id: TerminalTabID) {
     guard let index = tabs.firstIndex(where: { $0.id == id }) else { return }
     tabs.remove(at: index)
+    clearStalePendingRename()
     guard selectedTabId == id else { return }
     if index > 0 {
       selectedTabId = tabs[index - 1].id
@@ -94,19 +96,37 @@ final class TerminalTabManager {
 
   func closeOthers(keeping id: TerminalTabID) {
     tabs = tabs.filter { $0.id == id }
+    clearStalePendingRename()
     selectedTabId = tabs.first?.id
   }
 
   func closeToRight(of id: TerminalTabID) {
     guard let index = tabs.firstIndex(where: { $0.id == id }) else { return }
     tabs = Array(tabs.prefix(index + 1))
+    clearStalePendingRename()
     if let selectedTabId, !tabs.contains(where: { $0.id == selectedTabId }) {
       self.selectedTabId = tabs.last?.id
     }
   }
 
+  func beginTabRename(_ id: TerminalTabID) {
+    guard tabs.contains(where: { $0.id == id && !$0.isTitleLocked }) else { return }
+    pendingRenameTabID = id
+  }
+
+  func consumePendingRename() {
+    pendingRenameTabID = nil
+  }
+
   func closeAll() {
     tabs.removeAll()
+    pendingRenameTabID = nil
     selectedTabId = nil
+  }
+
+  private func clearStalePendingRename() {
+    if let id = pendingRenameTabID, !tabs.contains(where: { $0.id == id }) {
+      pendingRenameTabID = nil
+    }
   }
 }
--- a/supacode/Features/Terminal/Models/TerminalTabManager.swift
+++ b/supacode/Features/Terminal/Models/TerminalTabManager.swift
@@ -5,9 +5,15 @@ import SupacodeSettingsShared
 @MainActor
 @Observable
 final class TerminalTabManager {
-  var tabs: [TerminalTabItem] = []
+  var tabs: [TerminalTabItem] = [] {
+    // Drops `editingTabID` when the edited tab disappears across any close path.
+    didSet {
+      guard let id = editingTabID, !tabs.contains(where: { $0.id == id }) else { return }
+      editingTabID = nil
+    }
+  }
   var selectedTabId: TerminalTabID?
-  var pendingRenameTabID: TerminalTabID?
+  private(set) var editingTabID: TerminalTabID?
 
   private static let logger = SupaLogger("TabManager")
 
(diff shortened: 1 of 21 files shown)


iMostfa (approved):

supacode/Features/Terminal/Models/TerminalTabManager.swift
```diff
@@ -49,9 +49,15 @@ final class TerminalTabManager {
   func updateTitle(_ id: TerminalTabID, title: String) {
     guard let index = tabs.firstIndex(where: { $0.id == id }) else { return }
     guard !tabs[index].isTitleLocked else { return }
+    guard tabs[index].customTitle == nil else { return }
```
sbertix: With this guard, `title` freezes the moment a rename lands, so the second field isn't carrying its weight today. If we drop it and let Ghostty keep writing to `title`, then `commitRename` could treat an empty submit as "clear `customTitle`" and `displayTitle ?? title` would snap back to the live shell title. That gives you a natural undo-rename gesture for free. Wdyt?

supacode/Features/Terminal/Models/TerminalTabManager.swift
```diff
@@ -49,9 +49,15 @@ final class TerminalTabManager {
...
     tabs[index].title = title
   }
 
+  func setCustomTitle(_ id: TerminalTabID, title: String?) {
+    guard let index = tabs.firstIndex(where: { $0.id == id }) else { return }
+    tabs[index].customTitle = title
```
sbertix: The trim + empty-check sits in `TerminalTabView.commitRename` today, which means every future entry point (the command palette being the immediate one, but also deeplinks or any reducer action later) has to re-implement the same validation, or skip it. Cause the manager is the source of truth, imho it should own the invariant, normalise empty/whitespace to `nil` here so any caller gets consistent behaviour.

supacode/Features/Terminal/Models/WorktreeTerminalState.swift
```diff
@@ -806,6 +806,7 @@ final class WorktreeTerminalState {
         TerminalLayoutSnapshot.TabSnapshot(
           id: tab.id.rawValue,
           title: tab.title,
+          customTitle: isBlockingScriptTab ? nil : tab.customTitle,
```
sbertix, line 808: The blocking-script scrub here is easy to regress silently. Could we add a direct test against `captureLayoutSnapshot` asserting a locked tab with a `customTitle` gets stripped? And while we're there, a round-trip test through `restoreFromSnapshot` (asserting `displayTitle == "foo"`) would close the loop. `makeLayoutSnapshot` in `WorktreeTerminalManagerTests.swift:984` already has the scaffolding.

supacode/Features/Terminal/TabBar/Views/TerminalTabView.swift
```diff
@@ -61,6 +67,28 @@ struct TerminalTabView: View {
       .animation(.easeInOut(duration: TerminalTabBarMetrics.hoverAnimationDuration), value: isHovering)
       .animation(.easeInOut(duration: 0.2), value: hasNotification)
       .padding(.trailing, TerminalTabBarMetrics.tabHorizontalPadding)
+      .opacity(isEditing ? 0 : 1)
+      .allowsHitTesting(!isEditing)
+    }
+    .overlay {
+      if isEditing {
+        TextField("", text: $editingTitle)
```
sbertix, line 79: NIT: the TextField has no `.accessibilityLabel`, so VoiceOver announces an unlabeled field. One-liner.

supacodeTests/TerminalTabManagerTests.swift
```diff
@@ -106,4 +106,35 @@ struct TerminalTabManagerTests {
...
   }
+
+  @Test func setCustomTitleOverridesDisplayTitle() {
+    let manager = TerminalTabManager()
+    let id = manager.createTab(title: "tab 1", icon: nil)
+    manager.setCustomTitle(id, title: "my name")
+    #expect(manager.tabs.first { $0.id == id }!.displayTitle == "my name")
+  }
+
+  @Test func setCustomTitleDoesNotLockTitle() {
```
sbertix, line 117: NIT: `setCustomTitle` never touches `isTitleLocked`, so this one can really only fail if someone adds a mutation on purpose. Fine as intent-doc, not really coverage.

sbertix (changes requested):
Works overall 💪, but a few things to sort before merge.

The guard in `updateTitle` means the second field isn't really earning its keep, the rename validation lives in the view instead of the manager, the blocking-script scrub at capture is untested, and the "Change Terminal Title" palette entry currently shows up but is a no-op, we should wire it in this PR.

### Command palette wiring

The "Change Terminal Title" entry shows up in the command palette today, cause we pipe Ghostty's native commands through via `.ghosttyCommand` (see `CommandPaletteFeature.swift` around line 462/571/672), but it's a no-op: it routes to Ghostty's own title setter, which `updateTitle` now ignores once a `customTitle` is set, and it has no path to `setCustomTitle` at all. So users see the action and nothing happens, which is worse than not having it.

Could we intercept that specific Ghostty command in this PR and route it to the same rename flow (either trigger the inline TextField overlay for the active tab, or open a mini input popover)? Otherwise we ship a broken entry point on day one.

supacode/Features/Terminal/Models/TerminalTabManager.swift
bitomule: Yes, I think you are right, it'll be a better UX if cleaning popups back to Ghostty title.

supacode/Features/Terminal/Models/TerminalTabManager.swift
bitomule: Totally right, missed this one. Fixed in the same commit as I moved new logic to tab removing any trace of logic in view.

supacode/Features/Terminal/Models/WorktreeTerminalState.swift
bitomule: Added, let me know what you think

supacode/Features/Terminal/TabBar/Views/TerminalTabView.swift
bitomule: Done.

supacodeTests/TerminalTabManagerTests.swift
bitomule: Yeah, agree it's low value but it documents that relationship between customTitle and isTitleLocked. Happy to remove it you think it's not helpfull.

sbertix (approved):
Thank you 🙇‍♂️

PR #271, opened 2026-04-23
feat: add local comms bridge conversation pane

minzique:
## Summary
- add a persisted per-worktree conversation store and bottom conversation pane in `WorktreeDetailView`
- add a harness-agnostic `supacode comms` CLI plus generic socket action-command plumbing
- fan out incoming comms messages through existing system notification settings and cover the reducer/socket parsing with tests

## Testing
- not run: `make build-app` (blocked in this shell because `mise`/`tuist` are unavailable: `mise: command not found`)


sbertix:
Hey 👋 this doesn't seem to build. Could you double check that and also expand on the actual feature/reasoning/implementation a bit?

PR #284, opened 2026-04-28
Tint toolbar and tab bar with terminal theme

### LINKED ISSUE #281: Feature request: top bar follow terminal theme
Currently we have option to "Sync with terminal", to make terminal colors follow supacode. But I want it to be other way around actually! I have a pretty terminal theme and want that supacode used terminal backgdound as color for tabs and top bar, which are currently just gray.

mchernigin:
Closes #281 

I tried to do it myself. Here's how it looked before:
image

And after:
image

### DIFF
--- a/supacode/Features/Terminal/BusinessLogic/WorktreeTerminalManager.swift
+++ b/supacode/Features/Terminal/BusinessLogic/WorktreeTerminalManager.swift
@@ -1,3 +1,4 @@
+import AppKit
 import Foundation
 import Observation
 import Sharing
@@ -467,6 +468,14 @@ final class WorktreeTerminalManager {
 
   func surfaceBackgroundOpacity() -> Double {
     runtime.backgroundOpacity()
+  }
+
+  func surfaceBackgroundColor() -> NSColor {
+    runtime.backgroundColor()
+  }
+
+  func surfaceBackgroundColorScheme() -> ColorScheme {
+    runtime.backgroundColorScheme()
   }
 
   func unfocusedSplitOverlay() -> (fill: Color?, opacity: Double) {
--- a/supacode/Features/Terminal/TabBar/TerminalTabBarColors.swift
+++ b/supacode/Features/Terminal/TabBar/TerminalTabBarColors.swift
@@ -2,39 +2,39 @@ import AppKit
 import SwiftUI
 
 enum TerminalTabBarColors {
-  static var barBackground: Color {
-    Color(nsColor: .windowBackgroundColor)
-  }
-
-  static var activeTabBackground: Color {
-    Color(nsColor: .controlBackgroundColor)
-  }
-
-  static var hoveredTabBackground: Color {
-    Color(nsColor: .controlBackgroundColor).opacity(0.5)
-  }
-
   static var inactiveTabBackground: Color {
     .clear
   }
 
   static var activeText: Color {
-    Color(nsColor: .labelColor)
+    .primary
   }
 
   static var inactiveText: Color {
-    Color(nsColor: .secondaryLabelColor)
-  }
-
-  static var separator: Color {
-    Color(nsColor: .separatorColor)
+    .secondary
   }
 
   static var dropIndicator: Color {
     Color.accentColor
   }
+}
 
-  static var dirtyIndicator: Color {
-    Color(nsColor: .labelColor).opacity(0.6)
+private struct SurfaceChromeBackgroundColorKey: EnvironmentKey {
+  static let defaultValue = Color(nsColor: .windowBackgroundColor)
+}
+
+private struct SurfaceChromeColorSchemeKey: EnvironmentKey {
+  static let defaultValue = ColorScheme.dark
+}
+
+extension EnvironmentValues {
+  var surfaceChromeBackgroundColor: Color {
+    get { self[SurfaceChromeBackgroundColorKey.self] }
+    set { self[SurfaceChromeBackgroundColorKey.self] = newValue }
+  }
+
+  var surfaceChromeColorScheme: ColorScheme {
+    get { self[SurfaceChromeColorSchemeKey.self] }
+    set { self[SurfaceChromeColorSchemeKey.self] = newValue }
   }
 }
--- a/supacode/Features/Terminal/TabBar/Views/TerminalTabBackground.swift
+++ b/supacode/Features/Terminal/TabBar/Views/TerminalTabBackground.swift
@@ -8,12 +8,21 @@ struct TerminalTabBackground: View {
   var isHovering: Bool
   var tintColor: TerminalTabTintColor?
 
+  @Environment(\.surfaceChromeBackgroundColor)
+  private var surfaceChromeBackgroundColor
+  @Environment(\.surfaceChromeColorScheme)
+  private var surfaceChromeColorScheme
+
   var body: some View {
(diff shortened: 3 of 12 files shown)


supacode/Features/Terminal/Views/WorktreeTerminalTabsView.swift
```diff
@@ -17,6 +17,8 @@ struct WorktreeTerminalTabsView: View {
     let state = manager.state(for: worktree) { shouldRunSetupScript }
     let _ = configReloadCounter
     let unfocusedSplitOverlay = manager.unfocusedSplitOverlay()
+    let surfaceChromeBackgroundColor = Color(nsColor: manager.surfaceBackgroundColor())
```
sbertix, line 20: 🤔 Have we tried this with `background-opacity` set in the Ghostty config? The tints compose, so at 0.75 the terminal area ends up denser than the chrome and the bar/toolbar visually diverge from the terminal they're meant to match. Can we fix this?

supacode/Features/Terminal/TabBar/Views/TerminalTabBackground.swift
```diff
@@ -8,12 +8,21 @@ struct TerminalTabBackground: View {
...
+  private var surfaceChromeColorScheme
+
   var body: some View {
     ZStack(alignment: .top) {
       if isActive {
-        TerminalTabBarColors.activeTabBackground
+        Rectangle()
+          .fill(surfaceChromeBackgroundColor)
+        Rectangle()
+          .fill(chromeOverlayColor.opacity(0.14))
```
sbertix, line 22: Tabs stick out a lot now imho, the semi-transparent overlay on top of the chrome bg makes them very prominent.
Could we go plain instead? Active tab background = the chrome background (no overlay), inactive tab = `.saturation(0)` + reduced opacity (mirrors the unselected-surface treatment). Cleaner read, plus it side-steps the inactive-window contrast issue we'd otherwise need to solve separately. Wdyt?

supacode/Features/Terminal/TabBar/Views/TerminalTabBarView.swift
```diff
@@ -33,8 +33,8 @@ struct TerminalTabBarView: View {
       )
     }
     .frame(height: TerminalTabBarMetrics.barHeight)
+    .environment(\.colorScheme, surfaceChromeColorScheme)
```
sbertix, line 36: 🤔 This propagates through `.popover` and `.contextMenu` presentations too, cause this HStack is the structural ancestor of the `+` popover (`TerminalTabBarTrailingAccessories.swift:66`) and the per-tab right-click menu (`TerminalTabsRowView.swift:50`). With macOS dark + a light Ghostty theme the dropdown popover and context menu render light against the dark sidebar.
Could we re-inject the system colorScheme inside each presented content, or scope the override to the row container only?

supacode/Features/Terminal/Views/WorktreeTerminalTabsView.swift
```diff
@@ -17,6 +17,8 @@ struct WorktreeTerminalTabsView: View {
...
+    let surfaceChromeColorScheme = manager.surfaceBackgroundColorScheme()
```
sbertix, line 21: When the Ghostty config has no `background` set, `backgroundColorScheme()` materializes via `isLightColor` once and freezes the result. After a macOS Light/Dark toggle the bar fill swaps but the overlay tint and forced colorScheme stay frozen, so separators go invisible until something else mutates state.
Could we observe `NSApplication.didChangeAppearanceNotification` (or KVO `NSApp.effectiveAppearance`) and bump `configReloadCounter`? Wdyt?

supacode/Features/Terminal/TabBar/TerminalTabBarColors.swift
```diff
@@ -2,39 +2,39 @@ import AppKit
...
-    Color(nsColor: .separatorColor)
+    .secondary
   }
 
   static var dropIndicator: Color {
     Color.accentColor
   }
+}
+
+private struct SurfaceChromeBackgroundColorKey: EnvironmentKey {
```
sbertix, line 22: Could we collapse these into a single `SurfaceChromeAppearance` env key, imho? Store only the `Color` (or `NSColor`), expose `colorScheme` and `overlayTint` as computed properties.
Cause right now the two defaults already drift: `Color(nsColor: .windowBackgroundColor)` is system-adaptive but `ColorScheme.dark` is hard-coded, so in light mode an unset reader gets a near-white background paired with `.white` overlay (the scheme picks the tint via `chromeOverlayColor`).
Bonus: the collapse absorbs the duplicated `chromeOverlayColor` / `separatorOpacity` helpers across `TerminalTabBackground`, `TerminalTabCloseButtonBackground`, and `TerminalTabDivider`. Wdyt?

supacode/Features/Terminal/TabBar/Views/TerminalTabBackground.swift
```diff
@@ -27,10 +36,18 @@ struct TerminalTabBackground: View {
...
           Rectangle()
-            .fill(TerminalTabBarColors.separator)
+            .fill(chromeOverlayColor.opacity(separatorOpacity))
             .frame(height: 1)
         }
       }
     }
   }
+
+  private var chromeOverlayColor: Color {
```
sbertix, line 46: Tied to the `SurfaceChromeAppearance` collapse: I'd keep `.white`/`.black` here (not switch to `Color.primary`, cause that resolves to `NSColor.labelColor` with α≈0.85 in dark mode, which silently shifts our `.opacity(0.14)` overlay by ~15%).
But we could expose this once as a computed `overlayTint` on the struct, instead of duplicating the ternary across `TerminalTabBackground`, `TerminalTabCloseButtonBackground`, and `TerminalTabDivider`. Wdyt?

supacode/Features/Terminal/BusinessLogic/WorktreeTerminalManager.swift
```diff
@@ -469,6 +470,14 @@ final class WorktreeTerminalManager {
     runtime.backgroundOpacity()
   }
 
+  func surfaceBackgroundColor() -> NSColor {
```
sbertix, line 473: NIT, could we return SwiftUI `Color` directly instead of `NSColor`? Cause the only caller (`WorktreeTerminalTabsView.swift:20`) immediately wraps `Color(nsColor: …)`, and it'd let the view drop `import AppKit`. Plus it'd match `surfaceBackgroundColorScheme()` right below, which already returns the SwiftUI type.

sbertix (changes requested):
Nice direction overall 💪. Few things before this is ship-ready:

1. Doesn't compose with `background-opacity` in Ghostty (the tints stack, so the terminal ends up denser than the chrome).
2. Tabs read a bit too prominently with the new overlay imho, plain bg + desaturated/dimmed inactive tabs would be cleaner.
3. The `\.colorScheme` override leaks into popovers and context menus.
4. Chrome goes stale on macOS Light/Dark toggle when Ghostty has no `background` set.
5. The two new env keys can drift out of sync, would be nice to collapse into one `SurfaceChromeAppearance` value.

Also, we might want to default "Sync with Terminal" to on now imho. The terminal would now drive most of the UI, and I'm a bit worried that off-by-default plus odd Ghostty defaults could compromise the look on first launch. Wdyt?

PR #291, opened 2026-05-05
Support pasting/dropping images from clipboard (e.g. Preview.app)

muninn-huginn:
## Summary

Currently you can drag an **image file** from Finder into a terminal and the path is inserted, but copy/pasting an **image** from Preview.app (or any app that puts only image data on the pasteboard) does nothing — both the drag handler and `Cmd+V` silently drop image-only payloads.

This PR makes both flows work by materializing the pasteboard image to a temp PNG and inserting its path, the same way Finder file drags do.

### Changes

- `GhosttyRuntime.swift` — adds `NSPasteboard.writeImageToTempFile()` which extracts `public.png` (or converts `public.tiff` via `NSBitmapImageRep`) and writes it to `tmp/supacode-pasted-images/pasted-<uuid>.png`, returning the ghostty-escaped path.
- `GhosttySurfaceView.swift`
  - Adds `.png` / `.tiff` to `dropTypes` so `draggingEntered` no longer rejects image-only drags.
  - `performDragOperation` falls back to the new helper when there is no `URL` / `fileURL` / `string`.
  - Overrides `paste(_:)` to detect an image-only clipboard before falling through to Ghostty's `paste_from_clipboard` binding.

### Test plan

- [ ] Cmd+C an image in Preview → focus terminal → Cmd+V → file path is inserted.
- [ ] Drag an image out of Preview onto the terminal → file path is inserted.
- [ ] Plain text paste still works.
- [ ] Finder file drag still works.
- [ ] Drag a URL still works.

(I was unable to build locally due to a zig 0.15.2 / macOS 26.4 SDK toolchain issue unrelated to these changes — relying on CI.)

---
<a href="https://app.blacksmith.sh/supabitapp/codesmith/supacode/pr/291"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need.</sup>

- [ ] Let Codesmith autofix CI failures and bot reviews

### DIFF
--- a/supacode/Infrastructure/Ghostty/GhosttyRuntime.swift
+++ b/supacode/Infrastructure/Ghostty/GhosttyRuntime.swift
@@ -739,6 +739,40 @@ extension NSPasteboard {
     return string(forType: .string)
   }
 
+  /// Extracts image data from the pasteboard (e.g. when copied from Preview.app)
+  /// and writes it to a temp PNG file. Returns the escaped file path on success.
+  /// Used to bridge image clipboards / drags into terminals so CLI agents can read them as files.
+  func writeImageToTempFile() -> String? {
+    let pngType = NSPasteboard.PasteboardType("public.png")
+    let tiffType = NSPasteboard.PasteboardType("public.tiff")
+
+    let pngData: Data?
+    if let direct = data(forType: pngType) {
+      pngData = direct
+    } else if let tiff = data(forType: tiffType),
+      let rep = NSBitmapImageRep(data: tiff)
+    {
+      pngData = rep.representation(using: .png, properties: [:])
+    } else {
+      pngData = nil
+    }
+
+    guard let data = pngData else { return nil }
+
+    let dir = FileManager.default.temporaryDirectory.appendingPathComponent(
+      "supacode-pasted-images",
+      isDirectory: true
+    )
+    do {
+      try FileManager.default.createDirectory(at: dir, withIntermediateDirectories: true)
+      let url = dir.appendingPathComponent("pasted-\(UUID().uuidString).png")
+      try data.write(to: url)
+      return Self.ghosttyEscape(url.path)
+    } catch {
+      return nil
+    }
+  }
+
   static func ghostty(_ clipboard: ghostty_clipboard_e) -> NSPasteboard? {
     switch clipboard {
     case GHOSTTY_CLIPBOARD_STANDARD:
--- a/supacode/Infrastructure/Ghostty/GhosttySurfaceView.swift
+++ b/supacode/Infrastructure/Ghostty/GhosttySurfaceView.swift
@@ -1265,6 +1265,14 @@ final class GhosttySurfaceView: NSView, Identifiable {
   }
 
   @IBAction func paste(_ sender: Any?) {
+    let pb = NSPasteboard.general
+    let hasText =
+      pb.string(forType: .string) != nil
+      || (pb.readObjects(forClasses: [NSURL.self]) as? [URL])?.isEmpty == false
+    if !hasText, let imagePath = pb.writeImageToTempFile() {
+      insertText(imagePath, replacementRange: NSRange(location: 0, length: 0))
+      return
+    }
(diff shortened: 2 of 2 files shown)


sbertix:
Hey 👋 

This does not appear to work, per your test plan. Could you double check? 

Also, could you expand a bit more on the rationale, and what agent required this workaround?  
Main agents (I’m thinking Claude Code and Codex) support images by default with CTRL + V, and this, although small in terms of lines, feels like a big deviation from expected behavior. I’d rather see this resolved at the agent level, tbh, but I’m open to hearing more 💪

muninn-huginn:
When using Claude Code with Alacritty it works fine i.e. I can copy from the preview app in mac to the terminal, but when using in Supacode, it doesn't work.

I will keep exploring.

I will double check my tests, and update.

sbertix:
> When using Claude Code with Alacritty it works fine

What version of Claude Code/supacode are you on? Cause CTRL + V always worked for me in Claude Code/Codex + supacode. Even today 🤔 CMD + V does/might not (but that's not the shortcut you should use inside Claude Code for images either way)

khoi:
I think this is a good to have and is actually safe if we keep the precedence of clipboard handling as

file URLs -> strings -> image data (a new thing)

I don't think even Ghostty does this, Warp and Alacritty handles it fine.

sbertix:
> I think this is a good to have and is actually safe if we keep the precedence of clipboard handling as

I do 100% agree that this can be safe. And, despite siding more on the "neutral to have" vs "good", I'm not even against the idea in principle. 

I am more concerned about the reasoning behind adding it: cause this serves virtually no purpose for non-agent workflows (e.g., if it's in Preview it already has a path, even if you're pasting directly from the internet), and agents already handle this "natively" (⌃V), without needing to interfere on our side. And for those more niche agents which don't, it would definitely benefit them more to support it themselves directly (using their own opinionated paths, behavior, etc). 
So at that point it would be "adding it, just for the sake of adding it", imho. 

(It doesn't seem to work in Warp either, btw, or at least for me)

khoi:
I didnt know that the agent handle this with ctrl + v? Let me test

khoi:
Yep, seems like CTRL+V is a thing that agents handle natively. I don't think us handling CMD + V here is a bad default to have though (Warp does this, I just checked again).

I don't mind copying PNG data and it ends up being pasted as a /tmp/random-temp-path.png into my shell.

I think cmux can take this even further; if you're SSH-ing to a server, it automatically reuses the SSH connection and uploads the file, so you can drag and drop/copy-paste an image from the local machine to the remote coding agents. I think that's actually a great UX to have (but maybe the users should be able to opt out of it via a config)

I didn't know about the CTRL+V thing and expected CMD+V to work in the terminal. Maybe we should ship this, wdyt @sbertix ?

sbertix:
> Warp does this, I just checked again

You're right, this does work in **Warp** but **ONLY** for **agents** (I tested it on a regular terminal, and stopped when it did not show the path, as this proposal was trying to achieve, without moving further). But it only works for agents, not **because** they handle the data themselves, like this would do, but because they [**intercept** it **and commit the CTRL+V**, in its place](https://github.com/warpdotdev/warp/pull/9553) (for what I've gathered). 
This is something I would **100% be in favor to mirror** too. 

> I think cmux can take this even further; if you're SSH-ing to a server, it automatically reuses the SSH connection and uploads the file, so you can drag and drop/copy-paste an image from the local machine to the remote coding agents. I think that's actually a great UX to have (but maybe the users should be able to opt out of it via a config)

This would have very different implications, though. It doesn't feel in scope with this imho

khoi:
> This is something I would 100% be in favor to mirror too.

Let's do it

PR #301, opened 2026-05-06
Paint terminal tint at the window layer; keep chrome transparent

### LINKED ISSUE #281: Feature request: top bar follow terminal theme

sbertix:
## Summary

- Move terminal-theme tint to the window layer (alpha 1 when opaque, `withAlphaComponent(opacity)` + OS blur otherwise) so the toolbar gap, sidebar gap, tab-bar area, and terminal surface read as one seamless tinted region.
- Force Ghostty's surface to `background-opacity = 0` via a bundled override; user's intended opacity is snapshotted from a clone of the user-only config and applied at the window level so the tint never doubles up.
- Long-lived `WindowChromeObserver` (mounted at `ContentView`) keeps the window background tracking appearance / fullscreen / config / system Light-Dark flips even when no Ghostty surface is rendered (Empty / Loading / Archived / Multi-select states).
- Tab bar drops its own background fill: inactive tabs differentiate via `.opacity()` + `.saturation()` reduction with the colored stripe mirroring the same ramp on press/drag; overflow scroll cue uses a `.mask` instead of an overlay shadow; animations honor `@Environment(\\.accessibilityReduceMotion)`.
- New `SurfaceChromeAppearance` env value (carries `colorScheme` + `systemColorScheme`, derives `overlayTint` + `separatorOpacity`); `windowTintColorScheme(manager:)` injects it after capturing the system scheme above the override; `inheritSystemColorScheme()` restores system scheme inside popovers / context menus.

## Test plan

- [ ] Light system + dark terminal theme: terminal tint stays dark across toolbar gap, sidebar gap, behind tab bar; toolbar buttons + popovers + context menus render against system Light appearance.
- [ ] Dark system + light terminal theme: inverse — tint stays light, chrome controls render against system Dark.
- [ ] Toggle macOS Light/Dark while on Empty / Loading / Archived / Multi-select states (no terminal): window background updates correctly.
- [ ] Set `background-opacity = 0` in Ghostty config: window is fully transparent (just blur over wallpaper).
- [ ] Set `background-opacity = 0.7`: window paints terminal tint at 70% with OS blur behind.
- [ ] Live-edit Ghostty config (change theme / opacity): chrome refreshes without restart.
- [ ] Fullscreen toggle: titlebar contrast stays correct in fullscreen with light system + dark terminal.
- [ ] Hover, press, drag on tinted inactive tab: stripe + foreground content animate together.
- [ ] Inline-rename a tab: text is readable against the tinted bar (uses chrome scheme, not system).
- [ ] Command palette: opens against system appearance regardless of terminal theme.
- [ ] `make test` passes (1028 tests + 9 new for `SurfaceChromeAppearance` + `WindowAppearanceState`).

Closes #281
Closes #284


khoi:
I think it looks good. 

There is 1 problem I found where if you switch the system appearance and you have multiple splits, you sort of have to click the pane for it to settle to the final look 

https://github.com/user-attachments/assets/1344565b-7436-4071-a6eb-f5cf33e242aa

khoi (approved):
Looks good!

PR #308, opened 2026-05-11
Add per-worktree title and color customization

### LINKED ISSUE #300: Feature request: customize worktree display name (sidebar alias, decoupled from branch)
First off, Supacode has genuinely changed how I work. Running 8+ worktrees in parallel across Claude Code, Codex, and Pi as a single command center is the productivity setup I didn't know I needed. Thank you for building it.

Following the lovely [Customize Repository…](https://github.com/supabitapp/supacode/issues/275) sheet [@sbertix](https://github.com/sbertix) shipped recently, which lets us override a repo's sidebar name and color, I'd love the same affordance applied one level down, on individual worktrees.

### Why

My team's branch names encode the ticket ID by convention (`marvin/abc-432-link-payment-method-frontend`, `marvin/abc-9948-paywall-bounce-sms-frontend`, etc.). Those names are great for Git, GitHub, Graphite, and CI, but in Supacode's sidebar they all blur together as `marvin/...` and I lose the ability to scan at a glance.

The existing rename button in the worktree's `⋯` menu solves the wrong problem: it mutates the underlying git branch. For anyone using [Graphite](https://graphite.dev) stacks that's a non-starter. Renaming after `gt submit` orphans the PR (the GitHub `headRefName` is frozen), breaks parent/child pointers in the stack, and desyncs `gt restack`. Same story for any branch-name-keyed automation (ticket auto-link, CI branch protection, etc.).

Tab renaming ([#269](https://github.com/supabitapp/supacode/pull/269)) is great for the active terminal tab but doesn't change the sidebar entry I'm trying to scan.

### Proposal

Add a **Customize Worktree…** entry in each worktree row's `⋯` menu, mirroring the existing Customize Repository sheet:

- **Display name**: string, sidebar uses it when set, falls back to the branch name when empty.
- **Color**: optional, same palette as the repo customization.

### Implementation sketch (only if helpful)

Storage in `~/.supacode/settings.json` alongside `pinnedWorktreeIDs`:

```jsonc
{
  "worktreeOverrides": {
    "<worktree-id>": { "displayName": "Link payment frontend", "color": "blue" }
  }
}
```

Sidebar render: `overrides[worktreeId]?.displayName ?? branchName`. Cleanup hook on worktree delete drops the entry.

### Behavior notes

- Per-machine, not synced (matches repo customization today).
- Zero effect on git refs, branch names, tab titles, GitHub PRs, Graphite stacks, or any external integration.
- Pure UI override, safe by construction.

Happy to send a PR if you'd welcome one. Wanted to validate the shape first since the repo customization pattern is so close. Either way, thanks again for Supacode.

marvtub:
## TL;DR

Right-click any worktree in the sidebar → **Customize Worktree…** → give it a display title and/or a color tint. The original branch name and git ref are untouched, so Graphite stacks, GitHub PRs, and CI keep working — this is purely a sidebar override. The same Title / Color options now appear in the **New Worktree** dialog too, so you can name and tint a worktree the moment you create it.

Persists in `sidebar.json` next to the existing per-repo customization.

## Summary

- New **Customize Worktree…** sheet reachable from each worktree row's `⋯` menu, mirroring the existing Customize Repository flow. Sets an optional display title (used in the sidebar when non-empty, branch name otherwise) and an optional tint from the same `RepositoryColor` palette repos use.
- New Title / Color section in the **New Worktree** dialog, defaulting to empty / none — set them at creation, no extra step needed for the common case.
- Storage lands on `SidebarState.Item` (new optional `title` / `color` fields, lossy-decoded via `decodeIfPresent` / `encodeIfPresent` for backwards compat) so customization persists in `sidebar.json` next to the existing per-repo entries — no new settings file.
- Sidebar row, archive view, detail title, and toolbar notification popover all pick up the custom title (falling back to the branch name) and the picked tint; tint is suppressed under selection / highlight prominence so the row stays readable when chosen.
- Folder-kind repositories and the main worktree row are excluded from the menu entry (no override surface to render).
- Customization survives pin / unpin / archive transitions and a transient roster reload.

## Implementation notes

- Per-row `customTitle` / `customTint` live on `SidebarItemFeature.State` as mirrors of the bucketed `SidebarState.Item` (fanned in by `reconcileSidebarItems`); `SelectedWorktreeSlice` carries them so the detail title stays in lock-step with the sidebar row. `worktreeCustomization.save` participates in the `CacheInvalidations` system (`.all`) so the structure cache, selected-slice cache, and toolbar-notification cache all rebuild on save.
- `WorktreeCustomizationFeature` mirrors the canonical `RepositoryCustomizationFeature` sheet pattern (`BindableAction`, delegate, `.ifLet` presentation). The customization sheet and the New Worktree dialog both bind to the shared `ColorSwatchRow(color: $store.color)` directly — no per-feature `customColor` mirror state.
- Customization seeds onto `PendingWorktree.customization` when the New Worktree dialog submits, then transfers to the sidebar item once creation succeeds. The repo-reload pruner inside `applyRepositories` also matches pending rows to discovered worktrees by name (`progress.worktreeName`) and transfers customization onto the matched worktree's sidebar item before dropping the pending row — so a reload that races `createRandomWorktreeSucceeded` doesn't lose customization, and two concurrent pending creations in the same repo can't accidentally swap rows. Named pending rows only drop via exact name match; the count-based fallback never count-drops a named row. Regression-tested for both single and concurrent cases.
- Pin / unpin moves rows between buckets via `removeAnywhere` + `insert`; both helpers carry the bucketed `Item` forward so user-set `title` / `color` survive the move. `removeAnywhere` takes an explicit `preferring:` source-bucket order so a corrupted double-bucket pre-state surfaces the live row's payload, not a stale sibling's. The archive flow already carried the `Item` forward; pin / unpin and the toolbar notification group's name lookup are the matching paths.
- The reconcile pass preserves per-worktree title / color across transient roster reloads — same orphan-section pattern that protects per-repo customization in #276.
- `requestCustomizeWorktree` rejects folder-kind repos and main worktrees so a future deeplink or palette entry can't write customization that the row would never render. Same guard the menu entry uses; the context menu also hides the entry on pending rows since their in-flight `pendingID` isn't a real `worktree.id`.
- Sheets use the shared `.dismissSystemColorPanelOnDisappear()` modifier (also from #305), so a left-open `NSColorPanel` doesn't outlive the sheet.

## Test plan

- [x] `make check` (format + lint) clean for the changed files.
- [x] Manual: Create a worktree with a custom title + color from the **New Worktree** dialog → sidebar renders the customized row immediately; survives quit / relaunch.
- [x] Manual: Right-click a worktree → **Customize Worktree…** → set / clear title and color → sidebar updates and persists.
- [x] Manual: Archive a customized worktree → archive view shows the custom title and tint; unarchive restores the row with customization intact.
- [x] Manual: Pin a customized worktree → title and color survive the pin / unpin round-trip.
- [x] Manual: **Customize Worktree…** hidden on folder-kind repos, on main worktree rows, and on pending rows.
- [x] Fork CI green on the latest HEAD (build + test). Local `make build-app` / `make test` are blocked here by an unrelated `zig 0.15.2` + macOS 26 SDK linker issue, so verification ran through fork CI with a debug-artifact upload on a sibling branch — that temp step is dropped from the PR branch.

## Backwards compatibility

Existing `sidebar.json` files decode unchanged — both new `Item` fields use `decodeIfPresent`, defaulting to `nil`. No migration required, no schema bump needed.

Closes #300.


marvtub:
1-context-menu
2-customization-sheet
3-sidebar-result
4-new-worktree-dialog

Some screenshots I made during local smoke tests.

sbertix:
Apologies for not reviewing this yet. I'll try to get to it later today or tomorrow 💪 @marvtub

sbertix:
I took the liberty of swapping around some things in order not to drag this further 🙇‍♂️ 
Thank you sho much for this :muscle:

marvtub:
> I took the liberty of swapping around some things in order not to drag this further 🙇‍♂️ Thank you sho much for this 💪

This is awesome! Thanks so much for your help! Super excited to see this go live 🎉

PR #313, opened 2026-05-12
Persist window position and size across sessions

### LINKED ISSUE #268: Window doesn't persist position and size on macOS between sessions on a multi display setup
When closing Supacode and reopening it, it defaults to position itself filling the whole space of the primary display, even when I positioned it to the secondary display.

I'd expect the window to pop up again where I had it before closing Supacode.

tamarazuk:
## Summary

Fixes #268 — the main window now remembers its position and size across app launches, including when placed on a secondary display.

## Problem

Supacode had no window frame persistence. Every launch reset the window to fill the primary display, which is especially annoying on multi-monitor setups.

## Solution

Set `NSWindow.setFrameAutosaveName("main")` on the main window in `WindowTabbingView.disallowTabbing()`, which already configures the main `NSWindow` (tabbing mode, identifier, delegate).

Once an autosave name is set, macOS automatically:
- Saves the window frame to `UserDefaults` on every move/resize
- Restores the frame on next launch, including the correct display

A guard (`frameAutosaveName.rawValue != WindowID.main`) ensures the call is idempotent across repeated `updateNSView` cycles.

## Testing

- Single display: window position/size persists across quit & relaunch
- Multi-display: window moved to secondary display stays there after relaunch
- Display disconnected: macOS constrains the window to a visible display (built-in behavior)

Note: `make build-app` could not run in CI-like env (no Xcode installation), but the change is a single standard NSWindow API call with no new imports.

### DIFF
--- a/supacode/App/WindowTabbingDisabler.swift
+++ b/supacode/App/WindowTabbingDisabler.swift
@@ -21,7 +21,7 @@ final class WindowTabbingView: NSView, NSWindowDelegate {
     guard let window else { return }
     window.tabbingMode = .disallowed
     window.identifier = NSUserInterfaceItemIdentifier(WindowID.main)
-    if window.frameAutosaveName.rawValue != WindowID.main {
+    if window.frameAutosaveName != WindowID.main {
       window.setFrameAutosaveName(WindowID.main)
     }
     if window.delegate !== self {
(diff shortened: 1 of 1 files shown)


supacode/App/WindowTabbingDisabler.swift
```diff
@@ -21,6 +21,12 @@ final class WindowTabbingView: NSView, NSWindowDelegate {
     guard let window else { return }
     window.tabbingMode = .disallowed
     window.identifier = NSUserInterfaceItemIdentifier(WindowID.main)
+    // Persist and restore the window frame (position + size) across launches,
+    // including when the window is on a secondary display. macOS handles the
+    // save/restore automatically once an autosave name is set.
```
sbertix:
```suggestion
```

It's pretty self explanatory.

sbertix (approved):
Thanks 💪

Issue #289, opened 2026-05-04
navigating the menu causes an unexpected removal of the options


Willyfrog:
Adding a gif that will make the issue clear
There is no clicking, just moving the mouse around

<img width="800" height="422" alt="Image" src="https://github.com/user-attachments/assets/abb978d2-0df2-41d0-84ef-5f35d4d8231e" />

PR #334, opened 2026-05-19
Persist terminal sessions across app launches with bundled zmx

### LINKED ISSUE #315: Feature request: Startup Script or auto reopen Claude sessions
I am using `supacode` with Claude code. If i restart the application (even with `Restore Terminal Layout` enabled) I need to then manually reopen my Claude code sessions. I created a script that matches the tab UUID to the Claude code session and when I type `cs` in the tab it reopens the last session used in that tab. It would be really nice if I was able to do one of the following:

- Auto-run my script when the application starts - tabs reopen
- Or even better: `Restore Terminal Layout` function would remember the session

Or maybe I am missing something. Thanks in advance.

sbertix:
## Summary

Wraps every Ghostty surface in a `zmx attach <session-id>` so the underlying shell, scrollback, and any running agents survive an app quit. zmx is bundled as a git submodule (v0.6.0), built from source via Zig on `make build-app`, and embedded at `Contents/Resources/zmx/zmx`. On next launch every surface re-attaches by its persisted UUID.

Two commits, layered:

**`e32f0206` Persist terminal sessions with bundled zmx**
- `ZmxClient` (attach / kill / ls), 5s subprocess timeout, drained pipes, sun_path budget probe with trailing-slash trim so a `TMPDIR` without slash doesn't desync.
- Sidebar onboarding card; slot priority pre-empts older highlight / nested-worktrees prompts.
- Close-surface focus invariant fix: `GhosttySurfaceView.shouldClaimFocus` reclaims firstResponder after split-tree rebuild without stealing from the command palette / inline rename. `TerminalSplitTreePane` wraps the per-tab projection read so SwiftUI invalidates on Cmd+D / Cmd+W.

**`372bf086` Quit confirmation, Terminate All menu, agent persistence**
- `ConfirmQuitMode` (auto / always / never). `auto` confirms only when active work would be lost (running scripts, mid-archive, busy agent). Legacy `confirmBeforeQuit: true` migrates to `.always`, `false` to `.never` so opt-out users don't silently get the dialog back.
- Three-button alert via `QuitConfirmationContext` matrix; copy adapts when blocking scripts are in flight. New `terminateSessionsOnQuit` toggle pre-selects the destructive path.
- File menu "Terminate All Terminal Sessions" gated on `state.hasAnyTerminalSurface` (a manager-emitted cached aggregate; the gate is a single Bool, never a `sidebarItems` iteration). Tears down tracked surfaces AND reaps daemon-hosted orphans, because zmx outlives the app.
- Agent presence (claude / codex / etc.) is captured into `SurfaceSnapshot.agents` on background / quit and restored on next launch. Off-main `kill(pid, 0)` drops dead pids, post-projection re-fan-out resurrects badges as soon as `surfaceIDs` lands (handles the `appLaunched` race). Multiple agents per surface supported.
- Per-tab close batches the zmx kill into one `withTaskGroup` + one `count=N` analytics event instead of N detached spawns. Launch reaper sweeps `supa-*` orphans from crashes / force-quits.

## Performance

The aggressive observation discipline from #289 / #323 is preserved: every aggregate (`hasAnyTerminalSurface`, agent presence, projection emit) is flip-detected and cached at the manager / reducer boundary so no view body iterates `sidebarItems` or any other `IdentifiedArrayOf`. Audited explicitly across the diff.

## Test plan

- [x] `make build-app` clean at both commits.
- [x] `make test` green at HEAD, 1361 tests including new coverage for socket-dir trim, layout / agent round-trip, multi-agent restore with mixed liveness, race-fix regression, legacy quit-mode migration.
- [x] `make check` (format + lint) clean.
- [ ] Manual: quit with `claude` running, relaunch, badge resurrects.
- [ ] Manual: Quit and Terminate empties `zmx ls` output afterwards.
- [ ] Manual: rapid Cmd+Tab does not write `layouts.json` more than once per ~1s.
- [ ] Manual: archive a worktree, confirm its entry vanishes from `~/.supacode/layouts.json`.
- [ ] Manual: open command palette while a split-tree rebuild fires, confirm focus stays in the palette.

Closes #315
Closes #206

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "Resources/git-wt"]
 	path = Resources/git-wt
 	url = https://github.com/khoi/git-wt.git
+[submodule "ThirdParty/zmx"]
+	path = ThirdParty/zmx
+	url = https://github.com/neurosnap/zmx
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ TUIST_INSTALL_STAMP := $(TUIST_GENERATION_STAMP_DIR)/.installed
 TUIST_DEVELOPMENT_GENERATION_STAMP := $(TUIST_GENERATION_STAMP_DIR)/development
 TUIST_SOURCE_GENERATION_STAMP := $(TUIST_GENERATION_STAMP_DIR)/none
 TUIST_SOURCE_RELEASE_GENERATION_STAMP := $(TUIST_GENERATION_STAMP_DIR)/none-release
-TUIST_GENERATION_INPUTS := Project.swift Workspace.swift Tuist.swift Tuist/Package.swift $(wildcard Tuist/Package.resolved) $(PROJECT_CONFIG_PATH) mise.toml scripts/build-ghostty.sh
+TUIST_GENERATION_INPUTS := Project.swift Workspace.swift Tuist.swift Tuist/Package.swift $(wildcard Tuist/Package.resolved) $(PROJECT_CONFIG_PATH) mise.toml scripts/build-ghostty.sh scripts/build-zmx.sh
 TUIST_GENERATE_CACHE_PROFILE ?= development
 TUIST_CACHE_CONFIGURATION ?= Debug
 VERSION ?=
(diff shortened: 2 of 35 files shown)


khoi:
big change I also tried to do it but couldn't get the UX right. Let me test this out

khoi:
tested and it works well so far, I'm thinking about how to communicate that supacode is using zmx under the hood for the tech savvy users who'd like to know what's going on

khoi:
CleanShot 2026-05-19 at 16 44 02@2x

maybe a short paragraph explaining it under this option? Somehting along the line of

`SC uses zmx to persists your sessions by default .... etc`

khoi (approved):
minor UX comment

PR #343, opened 2026-05-21
Route terminal key events to focused pane

### LINKED ISSUE #342: Ctrl-R is delivered to every split pane
## Summary

When a worktree has multiple split panes, pressing `Ctrl-R` can be delivered to every pane instead of only the active/focused pane. This is especially visible when multiple Claude Code sessions are running: one inactive pane can receive `Ctrl-R` and enter a stuck/search-like input state.

## Expected

Keyboard input should be routed to the single active Ghostty surface in the key window. Inactive split panes should not forward `keyDown`, `keyUp`, or modifier changes to Ghostty.

## Actual

`performKeyEquivalent` already checks that the surface is the window's first responder, but the lower-level `keyDown` / `keyUp` / `flagsChanged` paths did not have the same guard. If those paths are invoked for a stale/non-active surface, that surface can still forward the event to its PTY.

## Repro

1. Open a worktree with two split panes.
2. Run Claude Code or another TUI in both panes.
3. Focus one pane.
4. Press `Ctrl-R`.
5. Observe that the inactive pane can also receive `Ctrl-R` and become stuck / enter an unintended input mode.

## Proposed fix

Guard terminal keyboard input at `GhosttySurfaceView` so only `window.firstResponder === self` forwards keyboard events to Ghostty, while preserving windowless initialization/test paths.

sf-jin-ku:
## Summary

- Fixes #342.
- Add a first-responder gate before `GhosttySurfaceView` forwards `keyDown`, `keyUp`, or `flagsChanged` events to Ghostty.
- Keep windowless initialization/test paths permissive, but prevent inactive split panes from receiving terminal keyboard input.
- Add focused regression coverage for the routing predicate.

## Test plan

- [x] `xcodebuild test -quiet -workspace /private/tmp/supacode-ctrl-r-focused-pane/supacode.xcworkspace -scheme supacode -destination "platform=macOS" -only-testing:supacodeTests/GhosttySurfaceViewTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY="" -skipMacroValidation`
- [x] `make -C /private/tmp/supacode-ctrl-r-focused-pane build-app`

### DIFF
--- a/supacode/Infrastructure/Ghostty/GhosttySurfaceView.swift
+++ b/supacode/Infrastructure/Ghostty/GhosttySurfaceView.swift
@@ -640,6 +640,7 @@ final class GhosttySurfaceView: NSView, Identifiable {
   }
 
   override func keyDown(with event: NSEvent) {
+    guard acceptsKeyboardInput(for: event) else { return }
     guard let surface else {
       interpretKeyEvents([event])
       return
@@ -685,11 +686,13 @@ final class GhosttySurfaceView: NSView, Identifiable {
   }
 
   override func keyUp(with event: NSEvent) {
+    guard acceptsKeyboardInput(for: event) else { return }
     if suppressKeyboardLayoutChangeKeyUp(event) { return }
     sendKey(action: GHOSTTY_ACTION_RELEASE, event: event)
   }
 
   override func flagsChanged(with event: NSEvent) {
+    guard acceptsKeyboardInput(for: event) else { return }
     let mod: UInt32
     switch event.keyCode {
     case 0x39: mod = GHOSTTY_MODS_CAPS.rawValue
@@ -870,7 +873,7 @@ final class GhosttySurfaceView: NSView, Identifiable {
 
   private func localEventKeyUp(_ event: NSEvent) -> NSEvent? {
     if !event.modifierFlags.contains(.command) { return event }
-    guard focused else { return event }
+    guard focused, acceptsKeyboardInput(for: event) else { return event }
     keyUp(with: event)
     return nil
   }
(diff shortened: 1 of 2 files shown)


supacode/Infrastructure/Ghostty/GhosttySurfaceView.swift
```diff
@@ -1032,6 +1035,24 @@ final class GhosttySurfaceView: NSView, Identifiable {
     Self.moveFocus(to: self)
   }
 
+  func acceptsKeyboardInput(for event: NSEvent) -> Bool {
+    guard let window else { return true }
```
sbertix, line 1039: Isn't this inconsistent with the whole assumption? Why should we accept the input if there's no window? 🤔

supacode/Infrastructure/Ghostty/GhosttySurfaceView.swift
```diff
@@ -1032,6 +1035,24 @@ final class GhosttySurfaceView: NSView, Identifiable {
...
+    return Self.acceptsKeyboardInput(
+      hasWindow: true,
+      eventWindowMatches: event.window.map { $0 === window } ?? true,
```
sbertix, line 1042: Synthesize events have `window` `nil`, so this is load-bearing and, if the issue described were real, this wouldn't fix it for those, right? Cause all surfaces would get the `nil` `event.window` and at that point `window.firstResponder` becomes the "tie-breaker", as it already is 🤔

supacode/Infrastructure/Ghostty/GhosttySurfaceView.swift
```diff
@@ -1032,6 +1035,24 @@ final class GhosttySurfaceView: NSView, Identifiable {
...
+      firstResponderIsView: window.firstResponder === self
+    )
+  }
+
+  static func acceptsKeyboardInput(
+    hasWindow: Bool,
+    eventWindowMatches: Bool,
+    firstResponderIsView: Bool
+  ) -> Bool {
+    guard hasWindow else { return true }
```
sbertix, line 1052: Same apparent inconsistency as above.

supacodeTests/GhosttySurfaceViewTests.swift
```diff
@@ -102,4 +102,35 @@ struct GhosttySurfaceViewTests {
...
+    #expect(
+      GhosttySurfaceView.acceptsKeyboardInput(
+        hasWindow: false,
+        eventWindowMatches: false,
+        firstResponderIsView: false
+      )
+    )
+  }
+
 }
```
sbertix, line 136: These are basically tautologies. We're safe to skip 💪

sbertix (changes requested):
Left a comment to the issue itself, too. Trying to narrow down the concerns a bit more.

PR #374, opened 2026-06-01
Fix Kiro CLI version detection to use kiro-cli instead of kiro

dlackty:
## Problem

Users with kiro-cli 2.x installed were seeing 'unsupported version' errors because Supacode was checking the old `kiro` binary (which returns 0.12.263) instead of the new `kiro-cli` binary (which returns 2.5.0).

The Kiro project renamed their binary from `kiro` to `kiro-cli` and bumped the major version to 2.x.

## Changes

- Updated `runKiroVersionCommand` to call `kiro-cli --version` instead of `kiro --version`
- Updated `supportedVersionPrefix` from `1.` to `2.` to match kiro-cli 2.x
- Updated all test fixtures to use kiro-cli 2.x version strings
- Updated log messages and comments to reference kiro-cli

## Testing

- ✅ All 17 KiroSettingsInstallerTests passed
- ✅ `make build-app` succeeded
- ✅ `make lint` and `make check` passed
- ✅ App runs successfully and correctly detects kiro-cli 2.5.0

## Files Modified

- `SupacodeSettingsShared/BusinessLogic/KiroSettingsInstaller.swift`
- `supacodeTests/KiroSettingsInstallerTests.swift`

### DIFF
--- a/SupacodeSettingsShared/BusinessLogic/KiroSettingsInstaller.swift
+++ b/SupacodeSettingsShared/BusinessLogic/KiroSettingsInstaller.swift
@@ -14,9 +14,9 @@ nonisolated struct KiroSettingsInstaller {
   /// defaults in `ensureDefaultAgentConfig` may no longer match upstream and would
   /// silently override a legitimately different config — gate on this prefix to
   /// fail loudly instead.
-  static let supportedVersionPrefix = "1."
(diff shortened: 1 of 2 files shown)


sbertix (approved):
Thanks 🙇‍♂️

PR #396, opened 2026-06-15
Paste clipboard image data as a temp file path

l0ng-ai:
## Problem

After copying an image (a screenshot, an image from a web page or an editor), pasting it into a terminal CLI like Claude Code with Cmd+V does nothing — you have to manually save the image to a file and copy its path.

Root cause: `getOpinionatedStringContents()` only recognizes *file URLs* and *plain text*. When the clipboard holds only raw image data (no file URL, no text), it returns `nil`, so the paste silently fails.

## Change

When the clipboard holds only image data, write it to a temp PNG and paste that file path (ghostty-escaped), so the CLI in the terminal can read the image by path. This matches iTerm2 / upstream Ghostty behavior.

- Precedence unchanged: **file URL > plain text > image-to-file**; text/file paste is unaffected.
- Image normalized to PNG (PNG passes through, TIFF is transcoded), written to `$TMPDIR/supacode-pasted-images/` with a UUID filename so repeated pastes don't collide.
- Each successful write prunes pasted images older than 24h, so the temp dir doesn't grow unbounded while recent paths stay valid for active terminal sessions.
- Write failures are logged via `SupaLogger`.

## Tests

`ClipboardPasteImageTests`: PNG pass-through, TIFF→PNG transcode, nil when no image, write path + content verification, unique filenames, stale-file pruning, end-to-end fallback to a path when only an image is present, and text taking precedence over image.

### DIFF
--- a/supacode/Infrastructure/Ghostty/GhosttyRuntime.swift
+++ b/supacode/Infrastructure/Ghostty/GhosttyRuntime.swift
@@ -794,6 +794,7 @@ extension NSPasteboard.PasteboardType {
 
 extension NSPasteboard {
   private static let ghosttyEscapeCharacters = "\\ ()[]{}<>\"'`!#$&;|*?\t"
+  private static let logger = SupaLogger("Clipboard")
 
   static func ghosttyEscape(_ str: String) -> String {
     var result = str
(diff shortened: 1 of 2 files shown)


sbertix:
This was already proposed and the overall goal is not do it this way, but doing like all other apps that support this, "forcing Cmd+V" to behave like the supported "Ctrl+V". If you're interested in implementing that, I'll for sure merge that
https://github.com/supabitapp/supacode/pull/291#issuecomment-4397402036

PR #409, opened 2026-06-18
Fix "Dismiss All" notifications never clearing the popover (#385)

### LINKED ISSUE #385: Dismiss All notifications does not work
When I hover over the notification icon (this is weird to be honest, I would expect having to click it) I get the dialog with the notifications the Dismiss All button does not seem to be doing anything, just hides the dialog, the notifications never disappear.

https://github.com/user-attachments/assets/7d3a394e-9149-4ef5-8b32-778660bb4cc5

webdevtodayjason:
## Problem

Fixes #385. Hovering the toolbar notification icon and clicking **Dismiss All** hides the popover but the notifications never actually disappear — they reappear next time you open it.

## Root cause

The toolbar popover renders from the per-row mirror `SidebarItemFeature.State.notifications`, which is fed by the row projection (`WorktreeRowProjection`) via `.terminalProjectionChanged`. That projection was only re-emitted through `onNotificationIndicatorChanged`, and the state-side helper (`emitNotificationIndicatorIfNeeded`) gated that callback on the **aggregate `hasUnseenNotification` flag flipping**.

So when you dismiss notifications you've already read (the flag is already `false`), the live `WorktreeTerminalState` cleared its array but **no projection was re-emitted** — the sidebar/popover mirror kept the stale entries. The same gap silently affected single dismiss, mark-as-read, and incoming notifications that arrive already-read on a focused surface.

## Fix

Fire on **any** notification-list mutation instead of only on the aggregate-flag flip. Both downstream effects already self-dedupe — `emitNotificationIndicatorCountIfNeeded()` on the count, and the manager's `emitProjection` on the whole `WorktreeRowProjection` value — so over-calling is free and can't cause storms. Renamed the helper to `emitNotificationStateChanged()` and dropped the now-dead `previousHasUnseen` locals at the six call sites.

## Test

Added `dismissAllReadNotificationsStillEmitsProjection` — seeds all-read notifications (so the unseen flag stays `false`), dismisses all, and asserts the projection callback still fires and the list clears. Fails before the change, passes after.

> Note: verified with swiftlint + the regression test's logic locally; the full `make build-app` / `make test` couldn't run on my machine (zig 0.15.2 can't link against the local macOS 26.5 SDK), so I'm relying on CI for the compile + test gate.

### DIFF
--- a/supacode/Features/Terminal/Models/WorktreeTerminalState.swift
+++ b/supacode/Features/Terminal/Models/WorktreeTerminalState.swift
@@ -906,17 +906,15 @@ final class WorktreeTerminalState {
   }
 
   func markAllNotificationsRead() {
-    let previousHasUnseen = hasUnseenNotification
     for index in notifications.indices {
       notifications[index].isRead = true
     }
     clearAllSurfaceUnseenFlags()
     emitAllTabProjections()
-    emitNotificationIndicatorIfNeeded(previousHasUnseen: previousHasUnseen)
+    emitNotificationStateChanged()
   }
 
   func markNotificationsRead(forSurfaceID surfaceID: UUID) {
-    let previousHasUnseen = hasUnseenNotification
     for index in notifications.indices where notifications[index].surfaceID == surfaceID {
       notifications[index].isRead = true
     }
@@ -924,12 +922,11 @@ final class WorktreeTerminalState {
     if let tabId = tabID(containing: surfaceID) {
       emitTabProjection(for: tabId)
     }
-    emitNotificationIndicatorIfNeeded(previousHasUnseen: previousHasUnseen)
+    emitNotificationStateChanged()
   }
 
   /// Marks a single notification as read, leaving others untouched.
   func markNotificationRead(id: WorktreeTerminalNotification.ID) {
-    let previousHasUnseen = hasUnseenNotification
     guard let index = notifications.firstIndex(where: { $0.id == id }) else { return }
     guard !notifications[index].isRead else { return }
     let surfaceID = notifications[index].surfaceID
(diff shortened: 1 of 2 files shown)


supacode/Features/Terminal/Models/WorktreeTerminalState.swift
```diff
@@ -1958,10 +1952,18 @@ final class WorktreeTerminalState {
...
+  /// Signals that the notification list or its read-state changed. Fires
+  /// unconditionally: the row projection (`currentProjection()`) carries the
+  /// full `notifications` array + per-item `isRead`, so the sidebar/popover
+  /// mirror must re-sync on *any* mutation, not only when the aggregate
+  /// `hasUnseenNotification` flag flips. Gating on the flag flip dropped
+  /// dismiss / mark-read of already-read notifications (the popover never
+  /// cleared — issue #385). Both downstream effects self-dedupe
+  /// (`emitNotificationIndicatorCountIfNeeded` on count,
+  /// `emitProjection` on the whole `WorktreeRowProjection` value), so
+  /// over-calling here costs nothing — do not re-add a flag gate.
```
sbertix:
```suggestion
/// `currentProjection()` already includes the full list and per-item `isRead`,
/// so the sidebar/popover must re-sync on every mutation, not just when
/// `hasUnseenNotification` flips. Gating here broke dismiss / mark-read of
/// already-read notifications (#385). Downstream emits self-dedupe, so keep
/// this ungated.```

sbertix (approved):

sbertix:
Thank you 🙇‍♂️

webdevtodayjason:
Applied your suggested wording for the doc comment — thanks for the review! 🙏

PR #410, opened 2026-06-18
Fall back to /bin/zsh for non-POSIX login shells (#100)

### LINKED ISSUE #100: Repository add fails when login shell is Nushell (SHELL=nu)
## Summary
When a user’s login shell is Nushell (`nu`), repository validation/add flows can fail with `not a git repository` even for valid git repos.

## User-visible problems
- Supacode reports `not a git repository` for valid repositories.
- Repository add/open flow is blocked for users whose shell is `nu`.
- Error message is misleading, making diagnosis difficult.

## Reproduction
1. Set login shell or `SHELL` env to Nushell (example: `/run/current-system/sw/bin/nu`).
2. Launch Supacode.
3. Try to add/open a valid git repository.
4. Observe failure with `not a git repository`.

## Expected
- Adding/opening a valid repo should succeed regardless of whether the user’s shell is `bash`, `zsh`, `fish`, `nu`, etc.
- If a shell is unsupported for login command execution, Supacode should fall back to a supported shell path and continue.

## Actual
- Supacode picks the shell from `SHELL`/passwd directly.
- Login command execution uses a POSIX-style invocation (`-l -c` plus POSIX command snippets).
- With `nu`, this invocation is incompatible, command execution fails, and repo detection surfaces as `not a git repository`.

## Suspected root cause
Shell selection and login command construction assume POSIX shells, but `SHELL` may point to a non-POSIX shell (Nushell).

## Notes
This is currently reproducible on setups that default to Nushell (commonly Nix/NixOS environments).

webdevtodayjason:
## Problem

Fixes #100. With a non-POSIX login shell (Nushell `nu`, common on Nix/NixOS, but also pwsh / elvish / xonsh / csh / tcsh), adding or opening a valid repository fails with a misleading `not a git repository`.

## Root cause

Every git / `wt` / `gh` command in the login-shell path is run as:

```
<userShell> -l -c "<POSIX rc-sourcing>; exec \"$@\"" -- <executable> <args...>
```

The `-c` snippet is POSIX (or fish) syntax. `ShellClient` only special-cased `fish` and `bash` and sent the zsh/POSIX `exec "$@"` form for everything else. A non-POSIX shell can't interpret that snippet, so the command never execs the real binary — and the failure surfaces upstream as `not a git repository`.

## Fix

Route the one-shot command path through a new `ShellClient.loginShellInvocation(userShell:)`. It keeps shells we know how to drive — `zsh`, `bash`, `fish`, `sh`, `dash`, `ksh` — and falls back to `/bin/zsh` for anything else, so the command still runs (as the reporter suggested).

The fallback is scoped to **command execution only**. The interactive terminal still spawns the user's real shell via `defaultShellPath()`, so a Nushell user keeps their Nushell terminal — only the internal git/wt/gh invocations are normalized.

It's an allowlist (default-deny), so unknown/future shells fall back safely rather than being assumed POSIX.

## Test

`ShellClientLoginShellTests` covers the POSIX shells (driven directly), fish and bash (own snippets), and the Nushell/pwsh/elvish/xonsh/csh fallback to `/bin/zsh`.

> Note: verified with swiftlint + the new test's logic locally; full `make build-app` / `make test` couldn't run on my machine (zig 0.15.2 can't link the local macOS 26.5 SDK), so I'm relying on CI for the compile + test gate.

### DIFF
--- a/SupacodeSettingsShared/Clients/Shell/ShellClient.swift
+++ b/SupacodeSettingsShared/Clients/Shell/ShellClient.swift
@@ -251,20 +251,13 @@ nonisolated private func collectOutput(
 }
 
 extension ShellClient {
-  /// Builds the `(shell, -c command)` pair for running a one-shot command
-  /// through a login shell.
-  ///
-  /// We can only drive shells whose `-l -c` accepts our POSIX (or fish) snippet.
-  /// A non-POSIX login shell — nushell, pwsh, elvish, xonsh, csh/tcsh — chokes
-  /// on the POSIX rc-sourcing + `exec "$@"`, which surfaced as a bogus
-  /// "not a git repository" on Nushell setups (issue #100). For those we fall
-  /// back to /bin/zsh so the command still runs.
-  ///
-  /// This fallback is scoped to the one-shot command path only; the interactive
-  /// terminal keeps spawning the user's real shell via `defaultShellPath()`.
-  /// ponytail: allowlist of shells we know how to drive; unknown → zsh.
+  /// Builds the `(shell, -c command)` pair for a one-shot login-shell command.
+  /// We only drive shells we have a correct rc snippet for — zsh, bash, fish.
+  /// Anything else (nushell, sh/dash/ksh, pwsh, …) falls back to /bin/zsh, which
+  /// can actually parse the snippet, so the command runs instead of failing
+  /// (issue #100). The interactive terminal still uses the user's real shell.
   nonisolated static func loginShellInvocation(userShell: URL) -> (shell: URL, command: String) {
-    let drivable: Set<String> = ["zsh", "bash", "fish", "sh", "dash", "ksh"]
+    let drivable: Set<String> = ["zsh", "bash", "fish"]
     let shell =
       drivable.contains(userShell.lastPathComponent)
       ? userShell : URL(fileURLWithPath: "/bin/zsh")
--- a/SupacodeSettingsShared/Clients/Shell/ShellClient.swift
+++ b/SupacodeSettingsShared/Clients/Shell/ShellClient.swift
@@ -78,8 +78,8 @@ extension ShellClient: DependencyKey {
       )
     },
     runLoginImpl: { executableURL, arguments, currentDirectoryURL, log in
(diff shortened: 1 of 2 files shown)


SupacodeSettingsShared/Clients/Shell/ShellClient.swift
```diff
@@ -250,14 +250,34 @@ nonisolated private func collectOutput(
...
+    let command: String
+    switch shell.lastPathComponent {
+    case "fish":
+      command = "test -f ~/.config/fish/config.fish; and source ~/.config/fish/config.fish >/dev/null 2>&1; exec $argv"
+    case "bash":
+      command = "[ -f ~/.bashrc ] && . ~/.bashrc >/dev/null 2>&1; exec \"$@\""
+    default:
+      command = "[ -f ~/.zshrc ] && . ~/.zshrc >/dev/null 2>&1; exec \"$@\""
+    }
+    return (shell, command)
```
sbertix: `sh`/`dash`/`ksh` are in `drivable`, so they run as themselves but still hit the `default` branch and source `~/.zshrc`. zsh-only syntax there is a parse error and the command just never runs.

Could we drop `sh`/`dash`/`ksh` from `drivable` so they fall back to `/bin/zsh` (which actually understands the snippet), or skip the `.zshrc` sourcing for non-zsh shells?

sbertix (changes requested):
Thank you 🙇‍♂️ one actual concern, since we're fixing this, and just some overall comment trimming would be nice.

webdevtodayjason:
Great catch — you're totally right. `sh`/`dash`/`ksh` ran as themselves but still got the zsh `.zshrc` snippet, so sourcing zsh syntax under them would blow up before `exec` ever fired. Exactly the kind of bug this PR is supposed to kill. 😅

Fixed: `drivable` is now just the shells we actually have a correct rc snippet for — `zsh`/`bash`/`fish` — and everything else (nushell, sh/dash/ksh, pwsh, …) falls back to `/bin/zsh`, which can parse it. Also trimmed the doc comment down, and expanded the test so sh/dash/ksh are asserted to fall back too.

Pushed — lmk if you want it any tighter! 🙏

sbertix (approved):

PR #411, opened 2026-06-18
Searchable base-ref picker in the New Worktree dialog (#387)

### LINKED ISSUE #387: Feature request: Search in Base ref in New worktree dialog
In the New Worktree dialog when selecting Base ref it would be a much better UX to be able to just start typing and filtering branches across local and remote branches. 

I always know at least part of the name but I have to remember if it is already local or just remote, if there is for example a jira ticket prefix in the name I have to check a few option to get to my branch. 

Lets say I want to branch of "core-refactoring" but the branch is "feature/jira-1234/core-refactoring" and I do not remember jira-1234 it takes me a while to find the branch in endless menus.

webdevtodayjason:
## What

Closes #387. The **Base ref** selector in the New Worktree dialog was a hierarchical `Menu` (Local submenu + per-remote submenus + nested namespace submenus). Finding a deeply-namespaced branch meant remembering its prefix and drilling submenus — e.g. to branch off `feature/jira-1234/core-refactoring` you had to recall `jira-1234` and whether it was local or remote.

## Change

Replace the nested menu with a popover that **filters a flat list of local + remote refs as you type**:

- Matches any fragment of the full ref (remote, namespace, or leaf), case-insensitive — so typing `core-refactoring` jumps straight to `feature/jira-1234/core-refactoring`.
- Keeps the **Auto** and default-branch (**Local**) quick picks pinned at the top.
- Filter field auto-focuses on open; selecting a ref sends the existing `baseRefSelected` action and dismisses.

The filtering is implemented as pure helpers on `BaseRefBranchMenu` (`allRefs()` / `refs(matching:)`), so the view is a thin renderer. The now-unused nested-menu helper views are removed.

## Tests

`BranchMenuFilterTests` covers the flatten, the deep-fragment match (#387's scenario), case-insensitivity across remotes, a namespace segment that's also a branch (`release` + `release/v2`), and the no-match case.

> Note on verification: filter logic is covered by the new unit tests; I confirmed the changed files lint clean. I couldn't run the app to eyeball the popover locally (zig 0.15.2 won't link the macOS 26.5 SDK on my machine), so a quick visual sanity-check of the popover sizing/placement would be appreciated.

### DIFF
--- a/supacode/Features/Repositories/BusinessLogic/BranchMenuNode.swift
+++ b/supacode/Features/Repositories/BusinessLogic/BranchMenuNode.swift
@@ -6,6 +6,9 @@ import SupacodeSettingsShared
 struct BaseRefBranchMenu: Equatable {
   var localBranches: [BranchMenuNode]
   var remotes: [Remote]
+  /// The default-branch quick pick, dropped from the Local submenu but still a
+  /// selectable ref the search must surface; nil when no such local branch exists.
+  var hoistedLocalBranch: String?
 
   struct Remote: Equatable, Identifiable {
     let name: String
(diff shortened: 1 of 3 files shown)


sbertix:
Ideally we'd keep both, because the menu is significantly more practical in larger teams/OSS where you don't necessarily know the name of the repo in advance. 
I was thinking more of having the text input inline and either a popover (if we can easily juggle keyboard arrows) or even just a set of rows (even just 1 row with fake "subrows" inside it) being added below it. Would you be interested in iterating on this?

webdevtodayjason:
Love it — yeah, keeping the menu totally makes sense. Browsing is the right move when you don't know the branch name up front, no argument here. 🙌

Just pushed an update that does exactly that: the hierarchical menu stays put for browsing, and there's now an inline filter field right next to it. As you type it drops the matching local + remote refs in as rows right below the field (went with your "set of rows" idea so there's zero keyboard-focus wrestling). Capped it at 8 with a little "keep typing to narrow" nudge so the dialog doesn't blow up.

Filtering still runs through the same tested `BaseRefBranchMenu.refs(matching:)` helper, so the menu and the search share one source of truth.

I went rows-first to keep it simple and shippable, but I'm totally down to layer arrow-key nav + the popover route on top next if you wanna get fancy. Lmk what you think! 🚀

davydog187:
@webdevtodayjason wanted to implement this myself until i found your PR. Dope!

sbertix (approved):
Improved the UX and UI a bit cause it needed some love, but kept everything else 💪 thanks 🙇‍♂️

PR #412, opened 2026-06-18
Add OpenCode agent integration

### LINKED ISSUE #406: Add OpenCode agent integration (hooks + presence + CLI skill)
Hi, love Supacode, using it daily with Claude, Codex and Metacode. Would you take a PR adding OpenCode?

OpenCode is the OSS agent CLI from SST. Hook shape is same as Claude Code: ~/.config/opencode/settings.json with top-level "hooks" object, events SessionStart, UserPromptSubmit, PreToolUse, PostToolUse, Notification, Stop, SessionEnd. Meta's internal fork MetaCode uses the same path, so this covers both without any Meta branding upstream.

My plan is boring copy-paste of the Claude integration, no new architecture.

Two quick questions before I clean it up:
1. OK as first-class agent alongside Claude/Codex/Kiro/Pi, or prefer behind flag?
2. Global ~/.config/opencode/settings.json only for now like Claude does, skipping project-local .opencode.json, reasonable?

If yes I'll open the PR with tests passing and formatted. Thanks for building this.

ashwathraop:
Closes #406

Adds OpenCode (github.com/anomalyco/opencode) as a first-class agent alongside Claude/Codex/Kiro/Pi — presence shimmer, awaiting-input, and the Supacode CLI skill.

**Approach**
OpenCode does not use a Claude-style `hooks` block. Its config is `opencode.json` (no `hooks` key) and its extensibility surface is JS/TS plugins auto-loaded from `~/.config/opencode/plugins/`. So presence is delivered by a generated plugin that runs the same guarded OSC 3008 command every other agent uses, single-sourced from `AgentHookSettingsCommand`:

| Supacode state | OpenCode trigger |
|---|---|
| session_start | plugin init |
| busy | `tool.execute.before` |
| idle | `tool.execute.after` / `session.idle` event |
| awaiting_input | `permission.ask` hook |
| busy (resume) | `permission.replied` event |
| session_end + idle | `dispose` hook |

The notification leg is intentionally omitted: OpenCode's `session.idle` event carries no assistant text (unlike Claude's Stop stdin), so there is nothing to forward without fabricating a body.

**Changes**
* `SkillAgent` — new `opencode` case (config `.config/opencode`, display "OpenCode", asset `opencode-mark`); wired through `integrationSubtitle`.
* `OpenCodePluginContent` / `OpenCodePluginInstaller` — generate and install the plugin; idempotent, reports `.outdated` only for a Supacode-owned stale plugin, never overwrites a user file of the same name.
* `AgentIntegrationFactory` — wire the opencode integration (plugin + CLI skill).
* `CLISkillContent` / `CLISkillInstaller` — opencode SKILL.md installed to `~/.config/opencode/skills/supacode-cli/` (OpenCode's global skills path).
* `opencode-mark` — monochrome glyph rendered as a template, matching the pi mark.
* Tests — `OpenCodePluginInstallerTests` + opencode byte snapshots.

**Verification**
* swiftc -parse clean; swiftlint --strict clean on changed files.
* OSC mechanism verified end-to-end in a shell (JSON-encode → `sh -c` → emits the OSC, guarded by SUPACODE_SURFACE_ID).
* TODO before merge: `make build-app && make test && make check` on macOS.

**Known follow-ups (deferred)**
* Text-only turns don't shimmer (busy is tool-triggered); could wire busy to the `chat.message` hook.
* Hooks `await` the OSC subprocess; could fire-and-forget or write to /dev/tty natively to drop per-tool latency.


sbertix (approved):
Thank you 🙇‍♂️

Issue #282, opened 2026-04-28
Supacode shortcut key usage issue


qianchencc:
When I need to use Ctrl-^. to exit Mosh in the Supacode terminal window, Ctrl is occupied by Supacode. It's only recognized as the Supacode window switching hotkey, not the hotkey I use to exit Mosh within the window. This annoys me, and I have to close the window.

I've already turned off all hotkeys in the settings, but it hasn't solved my problem.

Supacode is a great tool; it has provided me with many conveniences. Hopefully, this issue can be resolved soon.

sbertix:
Hey 👋 

`Ctrl + 6` (no shift)  + `.` (and disabling only "Select Worktree 6") doesn't work either?
It might require to be on the latest version to be correctly delegated to terminal.

PR #416, opened 2026-06-18
fix(shortcuts): honor overrides for terminal worktree-selection chords

### LINKED ISSUE #414: Hardcoded ctrl+0..9 tab switching in the terminal cannot be disabled or remapped, and blocks terminal key chords
### Summary

The terminal captures `ctrl` plus a number key (0 through 9) for tab switching (`goto_tab`). These bindings are hardcoded and are not affected by the shortcut settings. Disabling every shortcut in Settings, or remapping the "Select Worktree" shortcuts, has no effect on them. As a result, `ctrl+<digit>` chords never reach the running terminal program, which breaks things like the Mosh escape sequence `ctrl+^` (which is `ctrl+6`) and any shell or TUI binding that uses `ctrl` with a number.

### Environment

- Supacode: v0.10.2 (latest release at the time of writing)
- macOS: 26.5.1 (build 25F80)
- Reproduced on a standard US QWERTY layout

### Steps to reproduce

1. Open Settings, Shortcuts, and disable every shortcut (or at least all "Select Worktree" entries). You can confirm that `~/.supacode/settings.json` shows every entry under `global.shortcutOverrides` with `"isEnabled": false`.
2. Restart Supacode.
3. Focus a terminal and press `ctrl+6` (or any `ctrl` plus a digit).

### Actual behavior

`ctrl+<digit>` switches the tab instead of being delivered to the terminal. This happens even though all shortcuts are disabled, and it persists across restarts. The same behavior makes `ctrl+^` unusable as the Mosh escape, as reported earlier.

### Expected behavior

When the matching shortcut is disabled in Settings, `ctrl+<digit>` should pass through to the terminal. When it is remapped to a different key, the tab switch binding should move with it. The terminal tab switch bindings should respect the same override system as the other shortcuts.

### Root cause

In `SupacodeSettingsShared/App/AppShortcuts.swift`, the tab selection bindings are a static list that is appended unconditionally:

```swift
public static let tabSelectionGhosttyKeybindArguments: [String] = tabSelectionBindings.flatMap { binding in
  [
    "--keybind=ctrl+\(binding.unicode)=goto_tab:\(binding.tabIndex)",
    "--keybind=ctrl+\(binding.physical)=goto_tab:\(binding.tabIndex)",
  ]
}

public static func ghosttyCLIKeybindArguments(from overrides: [AppShortcutID: AppShortcutOverride]) -> [String] {
  let effectiveShortcuts = all.compactMap { $0.effective(from: overrides) }
  return effectiveShortcuts.map(\.ghosttyUnbindArgument) + tabSelectionGhosttyKeybindArguments
}
```

`tabSelectionGhosttyKeybindArguments` never reads `shortcutOverrides`. The per-shortcut `unbind` arguments derived from the overrides come first, and the unconditional `goto_tab` bindings are appended after them, so they take effect no matter what the user configured. The `selectWorktree` app shortcuts do flow through `effective(from:)`, which is why it looks like they should be configurable, but the separate hardcoded list wins in the terminal.

### Impact

Any `ctrl` plus number chord is unavailable in the terminal, with no way to opt out. Concrete examples are the Mosh escape `ctrl+^`, readline and shell bindings, and TUI apps that map `ctrl+<digit>`.

### Suggested fix

Route the tab selection bindings through the same `effective(from: overrides)` path as the other shortcuts, so that disabling emits `unbind` and remapping moves the binding. Alternatively, expose explicit shortcut entries for tab or worktree selection that the existing override UI can control, or add a dedicated toggle to release `ctrl+<digit>` to the terminal.

### Related issues

- #282 describes the same problem from a user's perspective (Ctrl captured in the terminal, Mosh `ctrl+^` unusable) and was closed without addressing the hardcoded bindings. I believe it would be resolved by this change.
- #182 (unusable worktree shortcuts on AZERTY) looks related, since the hardcoded bindings rely on both the unicode digit and the physical `digit_N` key and cannot be remapped. Making these configurable would likely help there too.
- #105 added the keybinding override system, but the terminal tab selection bindings were never brought into it.

ProjectMakersDE:
## What
Worktree-selection chords (default `ctrl+1` through `ctrl+0`) can now be disabled or remapped from Settings, and the change takes effect inside the terminal, not just in the menu bar.

## Why
`AppShortcuts.ghosttyCLIKeybindArguments(from:)` appended a fixed `tabSelectionGhosttyKeybindArguments` list that bound `ctrl+<digit>` to `goto_tab` unconditionally, ignoring `shortcutOverrides`. Disabling or remapping "Select Worktree N" in Settings only updated the menu-bar side, so the terminal still captured `ctrl+<digit>`. That blocked terminal chords such as Mosh's `ctrl+^` (which is `ctrl+6`) with no way to opt out. Reported in #414, same root cause as #282.

## How
`tabSelectionGhosttyKeybindArguments(from:)` now derives each `goto_tab` binding from the worktree-selection shortcut's `effective(from: overrides)`:
- Disabled: no binding is emitted, so the chord reaches the terminal.
- Remapped: the binding moves to the chosen key, freeing the old chord.
- Default: output is byte-for-byte identical, including the physical `digit_N` variant for non-US layouts (now emitted only while the binding stays the default Control+digit).

No Settings UI change is needed: the "Select Worktree N" entries already exist in the Worktree Selection group; this only makes them effective in the terminal.

## Testing
Updated the existing keybind tests to the new signature and added tests for the disabled and remapped cases (both `tabSelectionGhosttyKeybindArguments(from:)` and `ghosttyCLIKeybindArguments(from:)`).

I could not run the full build locally: our machine is on the latest Xcode 26.5 / Swift 6.3.2 and the macOS 26.5 SDK, which does not match the toolchain this project pins (the GhosttyKit zig build and some pinned dependencies do not compile against it), which is why this is not fully verified end to end on our side. The change is small and self-contained, and I verified the keybind-generation logic by compiling the affected source files in isolation, so I expect it resolves the issue. Please confirm via CI.

Fixes #414


sbertix (approved):
Thanks 🙇‍♂️

PR #432, opened 2026-06-20
Add Show Next/Previous Tab to the Window menu (#418)

### LINKED ISSUE #418: Feature request: Put "Show Next Tab" and "Show Previous Tab" in the application menu
I was looking for a way to quickly going from one tab to the next/previous and was sad when I couldn't find it anywhere.

But then I tried the keybinding that works in Ghostty, and that also works in Supacode 🎉

So if you could add "Show Next Tab" and "Show Previous Tab" to the "Window" menu item, then it will be possible to override the key-bindings using the macOS settings app.

webdevtodayjason:
## What

Closes #418. Tab switching already works via Ghostty's `next_tab` / `previous_tab` bindings, but there were no menu items for it — so it wasn't discoverable, and (as the issue notes) you couldn't rebind it from **System Settings ▸ Keyboard ▸ Shortcuts**, which keys off menu titles.

## Change

Adds **Show Next Tab** / **Show Previous Tab** to the **Window** menu (the macOS-conventional home for tab navigation).

Both route through the existing `performBindingAction` path with the `next_tab` / `previous_tab` binding strings — the *same* path Ghostty keybindings use — so any custom user bindings are honored, and there's no new tab-selection logic (the existing `handleGotoTabRequest` does the work).

Wiring mirrors the existing `splitTerminal` action end-to-end:
- two `AppFeature` actions (`showNextTab` / `showPreviousTab`)
- `FocusedAction`s published from `WorktreeDetailView`, gated on an active worktree
- menu items in `WindowCommands` with the Ghostty shortcut shown

## Test

`AppFeatureTabNavigationTests` covers both directions forwarding the correct binding and the no-selection no-op.

> Note: verified with swiftlint + the new tests' logic locally; I couldn't run the full build on my machine (zig 0.15.2 won't link the macOS 26.5 SDK), so I'm relying on CI for the compile + test gate. A quick check that the items land in the Window menu would be appreciated.


webdevtodayjason:
Refreshed: rebased onto current main (was 115 commits behind) and fixed the CI failure — the June `switch must be exhaustive` error came from main's new `affectsWorktreeMenuSnapshot` compile-time gate (added after this PR was written); `showNextTab`/`showPreviousTab` are now classified in its no-snapshot-input arm alongside `splitTerminal`. Verified the actions drive the same `selectTab` path the tab bar and Select Tab shortcut use, so this composes with the newer tab features (#712 hibernation wake included). swiftlint strict clean; swift-format no-op.

sbertix:
I just realized I didn’t explain why I’m withholding the merge. Apologies, I thought I had left a comment. 🙇‍♂️
We’re migrating away from the Ghostty-based tab system, so I’d rather see how this fits into the new one. I'm drafting the PR till then.

Issue #446, opened 2026-06-22
Enhancement: editor list omits Zed when only the Zed Preview channel is installed


mykhaliuk:
## What happens

The "Open in…" editor picker doesn't list **Zed** when only the **Zed Preview** channel is installed (no stable Zed). Users on the Preview channel can't select Zed at all.

## Steps to reproduce

1. Install only **Zed Preview** (`/Applications/Zed Preview.app`), not stable Zed.
2. Open the toolbar's "Open in…" / editor menu.
3. → Zed is missing from the list.

## Root cause

`OpenWorktreeAction` detects each app by a single hardcoded bundle id, and the Zed entry only checks the stable id `dev.zed.Zed`:

```swift
case .zed: "dev.zed.Zed"
```

`isInstalled` resolves it via `NSWorkspace.shared.urlForApplication(withBundleIdentifier:)`, which returns `nil` for a Preview-only install, so Zed is filtered out of `availableCases` (`menuOrder.filter(\.isInstalled)`).

Zed ships a distinct bundle id per release channel:

| Channel | Bundle id |
| --- | --- |
| Stable | `dev.zed.Zed` |
| Preview | `dev.zed.Zed-Preview` |
| Nightly | `dev.zed.Zed-Nightly` |
| Dev | `dev.zed.Zed-Dev` |

## Expected

Zed Preview is detected and selectable, consistent with how VS Code / VS Code Insiders / VSCodium are handled as separate entries. It opens via the same bundled CLI (`Contents/MacOS/cli`, present in the Preview app too).

## Environment

- macOS 26
- Zed Preview only (no stable Zed installed)

PR #447, opened 2026-06-22
fix(open-with): detect Zed Preview channel in editor list

mykhaliuk:
## Problem

The "Open in…" editor picker omits **Zed** when only the **Zed Preview** channel is installed. `OpenWorktreeAction.zed` detects installation via the single stable bundle id `dev.zed.Zed`, so `isInstalled` returns `false` for Preview-only setups (`dev.zed.Zed-Preview`) and Zed disappears from `availableCases`.

## Change

Adds a `.zedPreview` action mirroring `.zed`, following the existing VS Code / VS Code Insiders / VSCodium multi-channel convention:

- title **Zed Preview**, `settingsID` `zed-preview`, bundle id `dev.zed.Zed-Preview`
- same bundled-CLI open behavior (`Contents/MacOS/cli`, which the Preview app also ships) with workspace fallback
- listed right after `.zed` in `editorPriority`, so it flows into `menuOrder` / `defaultPriority`

Stable Zed and Zed Preview now surface independently, like the VS Code variants. The change is confined to `OpenWorktreeAction.swift` — the only place that switches over the enum.

## Scope

Preview only — the channel in #446. Nightly (`dev.zed.Zed-Nightly`) and Dev (`dev.zed.Zed-Dev`) can follow the same pattern later if wanted.

## Tests

- `zedPreviewUsesPreviewBundleIdentifierAndMirrorsZedOpenBehavior`
- `zedPreviewIsAnEditorListedAfterZed`

Build + tests run in CI (`ci.yml`).

### DIFF
--- a/SupacodeSettingsShared/Domain/OpenWorktreeAction.swift
+++ b/SupacodeSettingsShared/Domain/OpenWorktreeAction.swift
@@ -75,6 +75,7 @@ public enum OpenWorktreeAction: CaseIterable, Identifiable {
   case windsurf
   case xcode
   case zed
+  case zedPreview
 
   public var id: String { title }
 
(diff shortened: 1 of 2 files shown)


sbertix (approved):
Thank you 🙇‍♂️

PR #455, opened 2026-06-22
Add GitHub Copilot CLI as a coding agent

blacroix:
## Summary

Adds **GitHub Copilot CLI** (displayed as "Copilot CLI") to the `SkillAgent` agent-integration system, alongside Claude Code, Codex, Kiro, OpenCode, and Pi. Installing the integration wires up sidebar/tab agent-presence badges and notifications for Copilot, the same as the other agents.

## How it works

Copilot CLI auto-loads every JSON hook file under `~/.copilot/hooks/`, so Supacode writes and owns its own `~/.copilot/hooks/supacode.json` (the Pi/OpenCode "own-file" model) rather than merging into a shared settings object — no prune/append, no feature flag. The hooks emit the shared OSC 3008 presence signals, mapped to Copilot's lifecycle events (per GitHub's hooks docs):

| Copilot event | Signal |
|---|---|
| `sessionStart` | session_start |
| `userPromptSubmitted` / `preToolUse` / `postToolUse` | busy |
| `agentStop` | idle + done-notification |
| `sessionEnd` | session_end |
| `notification` (`permission_prompt` / `elicitation_dialog`) | awaiting_input + rich alert |

`pre`/`postToolUse` also recover the badge to busy after a permission prompt flips it to awaiting_input. The `notification` hook reads the payload and only acts on the "needs you" types so the done-notification (owned by `agentStop`) isn't duplicated.

## Changes

- `SkillAgent`: new `.copilot` case (`.copilot` dir, "Copilot CLI", `copilot-mark`)
- `CopilotHookSettings` / `CopilotHooksInstaller`: build + install/uninstall the owned `supacode.json` (sentinel ownership; only ever touches `supacode.json`, never sibling user hook files)
- `AgentPresenceOSC.emitNotifyShell` gains `readsStdin` so the branching `notification` hook can emit a notify from an already-captured payload
- Wired into `AgentIntegrationFactory`, `CLISkillInstaller`/`CLISkillContent`, and the Developer settings row
- `copilot-mark` asset (template-rendered, works in light + dark) + `CopilotHooksInstallerTests`

## Testing

- Unit tests pass: `CopilotHooksInstallerTests`, `AgentPresenceOSCTests`, `SettingsFeatureAgentIntegrationTests`.
- Verified at the hook level by capturing a live Copilot session: Copilot runs the installed hooks with the Supacode env present (`SUPACODE_SURFACE_ID` etc.) and the commands emit valid OSC 3008 to the surface tty (`write_to_tty=OK`). Presence events (`sessionStart`/`userPromptSubmitted`/`preToolUse`/`agentStop`) all fire.

### ⚠️ Not verified locally: end-to-end badges / notifications

I could **not** verify the rendered badge or notifications locally, because that requires a `GhosttyKit.xcframework` built with the OSC 3008 context-signal support (added June 2026). The local framework is from March and lacks the callback, and **rebuilding it isn't possible on my machine**:

- `make build-ghostty-xcframework` requires **zig 0.15.2** (ghostty's `requireZig` enforces this exact version), but zig 0.15.2 fails to link against the current **macOS 26.5 SDK** — a handful of re-exported libc symbols (`_sigaction`, `_waitpid`, `_realpath$DARWIN_EXTSN`, `_malloc_size`, `_posix_memalign`, `_sysctlbyname`, `_sigemptyset`) come back undefined (a known zig `.tbd` re-export bug). Same failure on the 15.4 / 26.4 / 26.5 SDKs.
- zig **0.16.0** (the only newer version available) both fails to compile ghostty (`process` has no `EnvMap`) and is rejected by `requireZig`.

So the presence-badge and notification **rendering** needs to be confirmed in CI / a build environment with a current GhosttyKit. The agent side (hooks fire, valid OSC emitted) and all unit tests are verified.

### Open question for review

Copilot's `notification` event did not fire for an in-CLI elicitation in my testing, so the permission/"needs you" alert depends on Copilot actually emitting `notification` events (its own setting). The presence states (busy/idle via the other hooks) are independent of that.


sbertix (approved):
Thank you 🙇‍♂️ just sorted the agents cases (so they read alphabetically) and adding some missing tests

PR #468, opened 2026-06-23
Self-diagnosing toolchain setup for macOS 26.4+

jeremybower:
## Why

A first-time / returning contributor on macOS 26.4+ (Tahoe) hits a chain of build failures, each surfacing only after the previous is fixed — and the worst one (the GhosttyKit Zig link failure) shows up as a ~200-line `undefined symbol: _malloc, _free, _sigaction, …` dump in `build_zcu.o` with no hint of the cause. The root cause: the pinned Zig (`0.15.2`, required exactly by ghostty) can't link the macOS 26.4+ SDK, which dropped the plain `arm64-macos` slice from `libSystem.tbd` ([ziglang/zig#31658](https://github.com/ziglang/zig/issues/31658), fixed only in Zig 0.16+). The fix is to build against Xcode 26.3 (ships the macOS 26.2 SDK).

This PR makes the prerequisites self-diagnosing and documents the setup, so contributors get a clear, actionable error the moment something is missing, and can keep multiple Xcodes side-by-side without a global `sudo xcode-select -s`.

## What (4 self-contained, individually-buildable commits)

1. **Pin dev tooling via mise** — `swift-format` (`spm:swiftlang/swift-format@602.0.0`, reproducible across contributors' Xcodes instead of the drifting toolchain `swift format`), `xcbeautify@3.2.1` (was piped but unpinned → pipefail failures), `swiftlint@0.64.0`. `make format` uses the pinned formatter; xcbeautify pipe tolerates absence.
2. **Auto-pin a Zig-linkable Xcode via `DEVELOPER_DIR`** — new `scripts/select-developer-dir.sh` scans candidate Xcodes and picks the first whose macOS SDK still has `arm64-macos` (probing with `xcrun --sdk macosx`, the form Zig uses). The ghostty/zmx build scripts and the Makefile build targets export it per-build — no global switch; `DEVELOPER_DIR=…` overrides. Adds the "Building on macOS 26.4+" docs.
3. **`make doctor`** — new `scripts/doctor.sh` checks mise-on-PATH, submodules, a Zig-linkable Xcode, Xcode license/first-launch, the Metal Toolchain, and the pinned tools, printing the exact fix for each failure. Build targets run it as a quiet order-only `preflight` (`SUPACODE_SKIP_PREFLIGHT=1` to skip).
4. **CI alignment** — the `setup-macos` action reuses the shared selector instead of its inline scan.

### Why no `patches/` entry
The link bug is in Zig's own self-hosted linker (`build_zcu.o`), not ghostty source, so the `patches/*.patch` mechanism (which only patches the ghostty submodule) can't fix it — and ghostty pins Zig to exactly `0.15.2`. The older-SDK + auto-`DEVELOPER_DIR` approach is the long-term fix until ghostty supports Zig 0.16+.

## Verification
- `make build-app` — full ghostty rebuild under the auto-pinned `DEVELOPER_DIR` + app → **Build Succeeded**
- `make format` → **zero `.swift` churn** (pinned formatter agrees with the tree)
- `make doctor` → all checks green; `--quiet` silent on success; failure path renders correct fixes
- `make lint` → clean


sbertix (approved):
Thanks 🙇‍♂️ trimmed some comments and did some minor adjustments

PR #480, opened 2026-06-24
fix: refuse to open a repo with duplicate worktree paths

### LINKED ISSUE #481: Adding a git repo with a corrupt worktree layout crashes the app on launch (persists across restarts)
### Summary
Adding a repository whose git config is corrupt can crash supacode on launch. Because the repo root is persisted, the crash recurs on every relaunch — there is no way back into the app without removing the persisted root.

### Root cause
A `Worktree`'s `id` is its working-directory path (`WorktreeLocation.id`, `supacode/Domain/RepositoryIdentity.swift:193`).

When a repository's git config is corrupt — e.g. a bare repo with `core.bare = false` and `core.worktree` pointing into one of its linked worktrees — `git` / `wt` report two worktrees at the **same path**, so the listing contains two `Worktree` values with the same `id`.

Both load paths feed that listing into `IdentifiedArray(uniqueElements:)`:
- `supacode/Features/Repositories/Reducer/RepositoriesFeature.swift:4180` (local, `wt ls --json`)
- `supacode/Features/Repositories/Reducer/RepositoriesFeature+Remote.swift:163` (remote, `git worktree list --porcelain`)

`IdentifiedArray.init(uniqueElements:)` (swift-identified-collections 1.1.1) has a **precondition that ids are unique** and traps (fatal, in release too) on a duplicate. The load runs on every launch → crash loop.

### Steps to reproduce
1. Have a repository whose worktree listing reports the same path twice. This happens when a bare-checkout repo has `core.bare = false` and a `core.worktree` redirect into a linked worktree; `wt ls --json` / `git worktree list` then return two entries with identical `path`.
2. Add that repository to supacode.
3. App crashes on load, and again on every relaunch.

Concrete trigger (observed `wt ls --json`):

```json
[{"branch":"main","path":"/repo/feature","head":"...","is_bare":false},
 {"branch":"feature","path":"/repo/feature","head":"...","is_bare":false},
 {"branch":"main","path":"/repo/main","head":"...","is_bare":false}]
```

(two entries share `/repo/feature`)

### Expected
The app should not crash. A repository that lists duplicate worktree paths is corrupt and should be surfaced as a load failure (error row), the same way missing/unreadable roots are, rather than crashing the process.

### Fix
PR #480 — detect a duplicate worktree id at the git/`wt` boundary and route the repo through the existing load-failure path instead of constructing the `IdentifiedArray`. Covers both the local and remote load paths.

### Environment
- supacode `main` @ e409e81a
- swift-identified-collections 1.1.1

tftio:
Closes #481

## Problem

Adding a git repository whose config is corrupt can crash the app on launch, and because the repo root is persisted the crash recurs on every relaunch — there's no way back into the app.

## Root cause

When a repository's git config is corrupt — e.g. a stale `core.worktree` redirect that resolves the bare root onto a linked worktree's directory — `git` / `wt` report **two worktrees at the same path**. Both load paths feed the listing straight into `IdentifiedArray(uniqueElements:)`, whose precondition **traps on duplicate ids** (a worktree's id is its working-directory path). The trap fires during repository load, which runs on every launch → crash loop.

Reproduced with a real bare-checkout repo whose `core.worktree` pointed into one of its linked worktrees; `wt ls --json` then returned two entries with the same `path`.

## Approach

Rather than crash — or silently dedupe and present a *guessed, possibly-wrong* worktree — **detect the duplicate and refuse to open the repository**, routing it through the existing load-failure path so the sidebar shows an error row explaining the repo may be corrupt. This validates at the boundary (the untrusted git/`wt` listing) and leaves the trusted-internal `IdentifiedArray(uniqueElements:)` construction unchanged — the same way missing/unreadable roots are already handled (`loadFailuresByID` → failure row).

Both the local (`wt ls --json`) and remote (`git worktree list --porcelain`) load paths get the guard.

## Changes

- `firstDuplicateWorktreeID(in:)` and `duplicateWorktreePathMessage(path:)` — shared by both load paths.
- Local: `loadRepositoriesData` refuses a repo with duplicate paths (`WorktreesFetchResult` with no worktrees + message); per-root fetch extracted into `worktreesFetchResult(for:gitClient:)`.
- Remote: `resolveRemoteRepository` keeps the placeholder and returns a `LoadFailure` for the same condition.

## Testing

- `firstDuplicateWorktreeIDFindsRepeatedPath` (unit) and `loadPersistedRepositoriesRefusesRepoWithDuplicateWorktreePaths` (end-to-end: corrupt repo → `loadFailuresByID` failure row, no crash).
- `swiftlint` passes on the changed files.
- Note: I could not run the full `xcodebuild`/`make test` suite locally — project generation fails building the bundled ghostty Zig dependency under zig 0.15.2 / macOS SDK 26.5 (`undefined symbol: _sigaction`), unrelated to this change. Relying on CI for the full build/test run.

## Note

Supersedes #479, which took a dedupe-and-recover approach. Refusing to open is more honest: it surfaces the corruption (which the user must repair) instead of silently showing a best-effort partial — and a repo reporting two worktrees at one path is never a healthy state to operate on.

### DIFF
--- a/supacode/Features/Repositories/Reducer/RepositoriesFeature+Remote.swift
+++ b/supacode/Features/Repositories/Reducer/RepositoriesFeature+Remote.swift
@@ -157,9 +157,8 @@ extension RepositoriesFeature {
       let loaded = try await client.gitWorktrees(for: rootURL)
       if !loaded.isEmpty {
         let worktrees = loaded.map { remoteWorktree(from: $0, host: host) }
-        // Same corruption guard as the local path: a healthy repo never lists
-        // two worktrees at one path. Refuse it (keep the placeholder + surface a
-        // failure) instead of trapping `IdentifiedArray(uniqueElements:)`.
+        // A duplicate path signals a corrupt repo; keep the placeholder and
+        // surface a failure instead of trapping `IdentifiedArray(uniqueElements:)`.
         if let duplicate = firstDuplicateWorktreeID(in: worktrees) {
           let failure = LoadFailure(
             rootID: repoID,
--- a/supacode/Features/Repositories/Reducer/RepositoriesFeature+Remote.swift
+++ b/supacode/Features/Repositories/Reducer/RepositoriesFeature+Remote.swift
@@ -156,11 +156,22 @@ extension RepositoriesFeature {
     do {
       let loaded = try await client.gitWorktrees(for: rootURL)
       if !loaded.isEmpty {
+        let worktrees = loaded.map { remoteWorktree(from: $0, host: host) }
+        // Same corruption guard as the local path: a healthy repo never lists
+        // two worktrees at one path. Refuse it (keep the placeholder + surface a
+        // failure) instead of trapping `IdentifiedArray(uniqueElements:)`.
+        if let duplicate = firstDuplicateWorktreeID(in: worktrees) {
+          let failure = LoadFailure(
+            rootID: repoID,
+            message: duplicateWorktreePathMessage(path: duplicate.rawValue)
+          )
+          return (remotePlaceholderRepository(host: host, remotePath: remotePath, repoID: repoID), failure)
+        }
         let repository = Repository(
           id: repoID,
           rootURL: rootURL,
           name: remoteRepositoryName(host: host, remotePath: remotePath),
-          worktrees: IdentifiedArray(uniqueElements: loaded.map { remoteWorktree(from: $0, host: host) }),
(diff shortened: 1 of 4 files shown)


sbertix (approved):
Thanks 🙇‍♂️ tidied some comments and reworked some tests

tftio:
Apologies for the slop issue; I hit "submit" too quick! Thanks for checking this out!

sbertix:
> Apologies for the slop issue;

Not at all. No worries 💪

Issue #441, opened 2026-06-21
GitHub CLI (gh) not detected: login-shell wrapper drops positional args (exec "$@")


mickeoL:
## Summary

When running supacode on an Intel Mac with gh installed via Homebrew, supacode fails to locate the GitHub CLI and throws `GithubCLIError.unavailable`. The code path that fails is `GithubCLIExecutableResolver.resolveExecutableURL`, which calls `locateExecutableURL` and attempts to run `/usr/bin/which gh` via both a non-login and a login shell. The login-shell invocation uses `exec "$@"` but when invoked the way supacode does, the positional arguments are not available inside the `-c` snippet and the lookup returns empty.

## Reproduction (from the user's environment)

1. gh is installed via Homebrew and is discoverable interactively:
   - `/usr/bin/which gh` or `which gh` prints `/usr/local/bin/gh`.
2. Running these directly works:
   - `/bin/zsh -l -c "/usr/bin/which gh"` → returns path.
3. But the exact command supacode constructs fails:
   - `/bin/zsh -l -c "[ -f ~/.zshrc ] && . ~/.zshrc >/dev/null 2>&1; exec \"$@\"" -- /usr/bin/which gh` → returns empty.

## Observed behavior in supacode

- The failing throw happens at `GithubCLIExecutableResolver.resolveExecutableURL` where both `locateExecutableURL` attempts return `nil` and the code throws `GithubCLIError.unavailable`.
- The non-login which invocation and the login-shell fallback behave differently when run the way supacode constructs them: direct `-c "/usr/bin/which gh"` works, but the `-c` snippet that sources `~/.zshrc` and then uses `exec "$@"` with `-- /usr/bin/which gh` does not make the positional arguments visible inside the snippet (echoing `$@` inside that snippet also yields empty).

## Root cause hypothesis

- The login-shell wrapper used by `ShellClient.runLoginImpl` constructs the shell command like this:
  - `execCommand = "[ -f ~/.zshrc ] && . ~/.zshrc >/dev/null 2>&1; exec \"$@\""`
  - final invocation: `/bin/zsh -l -c execCommand -- /usr/bin/which gh`
- For some reason in this invocation the `exec "$@"` form does not receive the `/usr/bin/which` + `gh` arguments as expected, so `which` is never executed and `locateExecutableURL` returns `nil`.
[...]

PR #482, opened 2026-06-24
Clear positional params before sourcing rc in login-shell probe (#477)

### LINKED ISSUE #477: GitHub CLI detection fails when ~/.zshrc sources a dual-mode script (login-shell probe leaks positional params)
## Summary

Settings → GitHub shows **"GitHub CLI not found"** even though `gh` is installed, on `PATH`, and authenticated. The root cause is in the login-shell probe: it leaks the target command into the shell's **positional parameters**, and any `~/.zshrc` that sources a *dual-mode* script (one that can be both sourced and executed, dispatching on `$1`) will hit that script's `exit` and kill the probe shell before `gh` runs.

This is environment-triggered, not user-misconfiguration: a perfectly normal `.zshrc` reproduces it.

- **Version:** 0.10.2 (142000)
- **OS:** macOS (Apple Silicon), Homebrew `gh` 2.95.0 at `/opt/homebrew/bin/gh` (+ `/usr/local/bin/gh` symlink), `gh auth status` ✓
- **Shell:** `/bin/zsh`

## Root cause

`GithubCLIExecutableResolver.locateExecutableURL` (Clients/Github/GithubCLIClient.swift) resolves `gh` via:

```
zsh -l -c '<command>' -- /usr/bin/which gh
```

where `<command>` is built by `ShellClient.posixLoginCommand` (SupacodeSettingsShared/Clients/Shell/ShellClient.swift):

```sh
__supacode_login_argv=("$@"); [ -f ~/.zshrc ] && . ~/.zshrc >/dev/null 2>&1; exec "${__supacode_login_argv[@]}"
```

Because the target (`/usr/bin/which gh`) is passed as positional parameters after `--`, the **positional parameters are set while `~/.zshrc` is sourced**. Every file that `.zshrc` sources inherits `$@` / `$1`.

If `.zshrc` sources a dual-mode script that dispatches on `$1` and `exit`s on an unknown subcommand, the sourced `exit` terminates the **whole probe shell** before `exec "${__supacode_login_argv[@]}"` runs. `stdout` is empty, `locateExecutableURL` returns `nil` for both the non-login and login attempts, and `resolveExecutableURL` throws `GithubCLIError.unavailable` → "GitHub CLI not found".

A very common trigger is [`junegunn/fzf-git.sh`](https://github.com/junegunn/fzf-git.sh), which many users source from `.zshrc`. It contains top-level `case "$1" in … *) exit 1 ;; esac`, so with `$1=/usr/bin/which` it exits during sourcing.

The existing `__supacode_login_argv` capture (added in #441 to survive `set --`) protects the *array*, but does not prevent a sourced script from `exit`ing on the leaked positionals.

## Reproduction

Minimal, no Supacode needed:

```sh
# A .zshrc that sources fzf-git.sh (or any dual-mode script that exits on unknown $1)
zsh -l -c 'print BEFORE; [ -f ~/.zshrc ] && . ~/.zshrc >/dev/null 2>&1; print AFTER' -- /usr/bin/which gh
# => prints only "BEFORE" (shell exits during source), exit 1

zsh -l -c 'print BEFORE; [ -f ~/.zshrc ] && . ~/.zshrc >/dev/null 2>&1; print AFTER'
# => prints "BEFORE" and "AFTER", exit 0
```

The full probe (`… ; exec "${__supacode_login_argv[@]}"` `-- /usr/bin/which gh`) therefore yields empty stdout.

## Proposed fix

Clear the positional parameters **before** sourcing the rc file. The command is already saved in `__supacode_login_argv`, so the exec is unaffected:

```sh
__supacode_login_argv=("$@"); set --; [ -f ~/.zshrc ] && . ~/.zshrc >/dev/null 2>&1; exec "${__supacode_login_argv[@]}"
```

Verified: with `set --` inserted, the unmodified `.zshrc` above completes sourcing and `/usr/bin/which gh` returns `/usr/local/bin/gh` (exit 0). The same applies to the `~/.bashrc` branch of `posixLoginCommand`. The fish branch (`exec $argv`) is unaffected since it doesn't reuse positionals, but clearing there too would be harmless.

## Workaround (user side)

Guard the dual-mode source so it only runs in interactive shells:

```sh
[[ -o interactive ]] && [ -f ~/fzf-git.sh/fzf-git.sh ] && source ~/fzf-git.sh/fzf-git.sh
```

dsenften:
Fixes #477.

## Problem

Settings → GitHub shows **"GitHub CLI not found"** even though `gh` is installed, on `PATH`, and authenticated.

The `gh` detection probe runs `zsh -l -c '<cmd>' -- /usr/bin/which gh`, passing the target as **positional parameters**. `ShellClient.posixLoginCommand` captured `$@` into a saved array (so a `set --` inside the rc couldn't wipe the exec, #441) but left the **live positionals set while sourcing the rc file**.

Any `~/.zshrc` that sources a *dual-mode* script dispatching on `$1` (e.g. [`junegunn/fzf-git.sh`](https://github.com/junegunn/fzf-git.sh), which contains a top-level `case "$1" in … *) exit 1 ;; esac`) then sees the probe's `/usr/bin/which`, hits that script's `exit`, and kills the probe shell **before `gh` is resolved**. `stdout` is empty, `locateExecutableURL` returns `nil`, and `resolveExecutableURL` throws `GithubCLIError.unavailable`.

This is environment-triggered, not user misconfiguration — a perfectly normal `.zshrc` reproduces it.

## Fix

Insert `set --` after the capture and before sourcing the rc file, for both the zsh and bash branches of `posixLoginCommand`:

```sh
__supacode_login_argv=("$@"); set --; [ -f ~/.zshrc ] && . ~/.zshrc >/dev/null 2>&1; exec "${__supacode_login_argv[@]}"
```

The `exec` reads from the saved array, so clearing the live positionals is safe. The fish branch (`exec $argv`) is unaffected.

## Verification

Behavioral repro with a real zsh dual-mode rc (script `exit`s on `$1`):

- **Without `set --`:** empty stdout — probe shell exits during sourcing before `exec`.
- **With `set --`:** the binary path is returned (exit 0).

Adds a regression test `zshAndBashClearPositionalsBeforeSourcingRc` asserting `capture → set -- → source` ordering in the generated snippet, alongside the existing #441 capture-ordering test.

> Note: the full Xcode test suite could not be run in my environment because the GhosttyKit Zig build fails to link (`undefined symbol: _waitpid`/`_fork`/… — a zig 0.15.2 × macOS 26.5 SDK libc toolchain issue, unrelated to this change). swiftlint passes clean on the changed files.

### DIFF
--- a/SupacodeSettingsShared/Clients/Shell/ShellClient.swift
+++ b/SupacodeSettingsShared/Clients/Shell/ShellClient.swift
@@ -288,13 +288,18 @@ extension ShellClient {
     return (shell, command)
   }
 
-  /// Builds the zsh/bash one-shot command: capture the positional parameters before sourcing the rc
-  /// file, then exec from the saved array. Sourcing shares `$@` with the caller, so an rc that resets
-  /// the positional parameters (e.g. `set --`) would otherwise wipe the command before `exec` (#441).
+  /// Builds the zsh/bash one-shot command: capture the positional parameters, clear them, then source
+  /// the rc file and exec from the saved array. Sourcing shares `$@` with the caller, so an rc that
+  /// resets the positionals (e.g. `set --`) would otherwise wipe the command before `exec` (#441).
+  /// Clearing `$@` with `set --` before sourcing also keeps the target command out of the rc's view:
+  /// a dual-mode script dispatching on `$1` (e.g. `fzf-git.sh`) would otherwise see the probe's
+  /// arguments, hit its own `exit`, and kill the probe shell before `exec` ran (#477). The exec reads
+  /// from the saved array, so clearing the live positionals is safe.
   nonisolated private static func posixLoginCommand(rcFile: String) -> String {
     let capture = "__supacode_login_argv=(\"$@\")"
+    let clear = "set --"
     let source = "[ -f \(rcFile) ] && . \(rcFile) >/dev/null 2>&1"
-    return "\(capture); \(source); exec \"${__supacode_login_argv[@]}\""
+    return "\(capture); \(clear); \(source); exec \"${__supacode_login_argv[@]}\""
   }
 }
 
--- a/supacodeTests/ShellClientLoginShellTests.swift
+++ b/supacodeTests/ShellClientLoginShellTests.swift
@@ -57,4 +57,24 @@ struct ShellClientLoginShellTests {
       #expect(command.contains("exec \"${__supacode_login_argv[@]}\""))
     }
   }
+
+  /// Regression for #477: after capturing the positional parameters, the zsh/bash snippet must clear
+  /// them (`set --`) BEFORE sourcing the rc file. The positionals otherwise leak into the rc, so a
+  /// dual-mode script dispatching on `$1` (e.g. `fzf-git.sh`) sees the probe's `/usr/bin/which gh`,
+  /// hits its own `exit`, and kills the probe shell before `gh` is ever resolved.
+  @Test func zshAndBashClearPositionalsBeforeSourcingRc() {
+    for path in ["/bin/zsh", "/bin/bash"] {
+      let command = ShellClient.loginShellInvocation(userShell: URL(fileURLWithPath: path)).command
+      guard let captureRange = command.range(of: "__supacode_login_argv=(\"$@\")"),
+        let clearRange = command.range(of: "set --"),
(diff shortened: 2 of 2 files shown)


sbertix (approved):
Thank you 🙇‍♂️

PR #496, opened 2026-06-26
feat: add support for IDEA EAP editor

Simek:
# Why

After installing Supacode spotted that my main editor is not visible in the selection list.

# How

Add IDEA EAP as a possible editor to open project with. It's Early Access Program version of JB IDEA (like nightly/beta), but the has different bundle identifier than the main IDE.

Ref:
* https://www.jetbrains.com/idea/nextversion/

# Test plan

Supacode has been build locally, all test and lint/format checks are passing.

# Preview

Screenshot 2026-06-26 at 13 44 05


sbertix (approved):
Thanks 🙇‍♂️

PR #497, opened 2026-06-26
feat: use user-defined order of repos in Settings sidebar

Simek:
# Why

While using Supacode for a while spotted that in Settings sidebar repositories are always ordered alphabetically. I think it would be nice to use the same, user-defined order like in main app sidebar.

# How

Read user-defined sidebar structure and pass ordered list when initializing SettingsView, reconstruct the SettingsRepositorySummary lists based on passed IDs.

# Test plan

Supacode has been build locally, all test and lint/format checks are passing. I have also made sure that after reordering repositories in main sidebar setting view always reflect and matches latest state.

# Preview

https://github.com/user-attachments/assets/0ffd447d-8a72-47b8-8907-d085a2bce65e


sbertix:
Thanks, but this is on purpose (similar to most settings panels showing apps, e.g., iOS, etc)

PR #505, opened 2026-06-27
Implement OSC 11 per-pane background tinting

sageframe-no-kaji:
## Summary

Implements OSC 11 per-pane background tinting with fixes for three correctness issues:

- **Layer ordering** — Ghostty's macOS renderer replaces `view.layer` with its IOSurface layer rather than adding a sublayer. A tint sublayer on that layer can composite above terminal contents. Fixed by splitting the surface into a container (`GhosttySurfaceView`) and render child (`GhosttyRenderView`): Ghostty owns the child layer, while the tint lives on the container layer behind it.

- **OSC kind clobber** — Color-change actions were stored as a single "latest color change", so OSC 10 (foreground), OSC 12 (cursor), or palette updates could overwrite OSC 11 background state and clear the pane tint. Fixed by routing background / foreground / cursor changes into separate state triples and ignoring palette/unknown kinds for pane tinting.

- **Blur proxy** — `backgroundTintOpacity` used `backgroundOpacity < 1.0` as a stand-in for blur, producing a washed-out 0.3 tint for configs like `background-opacity = 0.99` with no blur. Fixed by reading `background-blur` directly from Ghostty config via `ghostty_config_get`.

## Test plan

- [x] `make format`
- [x] `make lint`
- [x] `make test` — 128 suites passed; existing known-issue markers only
- [x] `make build-app`
- [x] Manual: `printf '\033]11;#882200\007'` applies a red pane background with text fully legible
- [x] Manual: OSC 10 / OSC 12 after OSC 11 do not clear the pane background
- [x] Manual: drag-and-drop, click-to-focus, ⌘C/⌘V, and window resize are unaffected

Screenshot 2026-06-27 at 8 37 35 AM


sbertix:
Sorry for keeping you waiting on this, btw. I was trying to find a way to reconcile the nature of our "fake window background color" without changing this, but I think your PR body explanation and the various commits trying to work around it (and they work) show we really need to address that first if we want to bring in OSC 11 to match Ghostty.

Let me fix that first and circle back. Sorry 🙇‍♂️

sageframe-no-kaji:
No worries at all—thanks for digging into it. Happy to hold while you sort the underlying piece. Ping me when you want me to rebase or tweak anything on the PR side.

sbertix:
I rewrote the background tinting completely in #570. It also supports OSC 11.

If you happen to try it out from the Tip channel (available in  ~30 min), and have any feedback or spot any bugs, feel free to reopen the PR targeting that 💪

Thank you 🙇‍♂️

PR #507, opened 2026-06-27
feat: add custom sidebar subtitle for worktrees

### LINKED ISSUE #491: Customize Appearance of sidebar subtitle as for sidebar title
Hi 👋🏻 

To reduce build time I am reusing my worktree so I named them (based on iOS simulator thigh to it, but it could be anything).

Is there a way to have branch instead of worktree/folder name (below custom name)?

Image

Would be useful (at least for me) to have the branch name used on that worktree instead of the worktree/folder's name.

Thanks!

Nominchuluunb:
Adds a user-editable subtitle field alongside the existing title and color in the worktree customization sheet. The custom subtitle overrides the auto-derived worktree name shown under each sidebar row, including inside the Pinned / Active highlight sections.

Also threads subtitle through pending worktree creation so a subtitle typed in the New Worktree appearance section survives creation.

Closes #491

### DIFF
--- a/supacode/Domain/PendingWorktree.swift
+++ b/supacode/Domain/PendingWorktree.swift
@@ -4,12 +4,19 @@ import SupacodeSettingsShared
 struct PendingWorktree: Identifiable, Hashable {
   /// Sidebar customization that travels with the in-flight creation. Reconcile
   /// reads this when materializing the pending row so a user-typed title /
-  /// color is visible while the worktree is being created; on success the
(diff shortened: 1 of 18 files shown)


sbertix:
Not a huge fan of adding extra granularity through subtitle customization, like I mentioned in the original issue comment. Also, the issue itself can be closed, as the workaround is more than good enough, so I would do the same to this. Sorry

PR #510, opened 2026-06-28
Add isolated dev build (make run-app-dev)

### LINKED ISSUE #635: Isolated dev build for safely testing state-touching changes
### What problem does this solve?

Working on Supacode means running an in-development build against your real `~/.supacode` and your actual worktrees. There's no isolation story: the dev build and your daily-driver install share the bundle id, the data directory, and every OS registration. I audited the codebase for what concretely goes wrong when two production instances run side by side (the only way to dogfood a change today without replacing your working install) — the failure modes are structural:

**Terminal sessions can kill each other.** Both instances restore the same `layouts.json`, so they derive identical `supa-<uuid>` zmx session names (`ZmxSessionID.make`) and attach to the *same* sessions — zmx broadcasts PTY output to all attached clients, so two windows silently share one shell. Closing a surface (or quitting with terminate-sessions) runs `zmx kill` unconditionally (`killZmxSessions`), destroying the other instance's live shell mid-command. This matches long-standing flakiness I've hit with two instances open.

**All persisted state is last-writer-wins.** The code says it outright — `LayoutsIncrementalWriter.swift`: *"a second Supacode instance writing the same file concurrently is a dev-only scenario and accepted as last-writer-wins."* The same whole-file pattern (atomic write, no lock, no re-read subscription) covers `settings.json`, `sidebar.json`, and repo-local `supacode.json`: a change made in the staler instance silently reverts the other's settings, pins, archives, and scripts. Both instances also share one UserDefaults domain, so live toggles bleed between them (sidebar nesting/grouping, remembered font size, palette recency, onboarding dismissals).

**OS-level routing picks one instance.** `supacode://` deeplinks — CLI actions, notification clicks, browser links — go to whichever instance LaunchServices chooses; the wrong one fails or acts in the wrong window. Each process starts its own Sparkle updater with auto-install, so one instance can replace the bundle the other is still executing from. And outside a Supacode terminal, every CLI command fails with *"Multiple Supacode instances found"* (`Dispatcher.resolveSocket`).

**Worktree lifecycle races.** Delete in instance A runs `git worktree remove --force --force` with no cross-process ownership check — ripping the cwd out from under B's live shells and scripts. Lifecycle guards and running-script registries are per-process, so a non-idempotent archive/delete script can run twice concurrently. Worktree creation is check-then-create (name collisions), and doubled HEAD watchers and PR polling double git churn and GitHub API burn.

**It also blocks two-ended (remote) feature work.** Anything remote — a binary proxying notifications back, an event stream, an SSH-side helper — is a protocol with the app on one end. Today the in-dev handlers must terminate in the app you daily-drive: a buggy handler pollutes your real notification stream, protocol migrations land mid-flight in prod, a crash loop takes down the app your agents run in. A dev build pairs the dev remote binary with the dev app end-to-end while prod stays on the released protocol — and makes version-skew testing runnable, since both pairs can be live at once.

In short: the codebase (reasonably!) assumes one instance per bundle id and data dir. That's exactly what makes dogfooding in-flight changes risky — today's safest workflow is "quit the app you use all day and hope nothing corrupts."

### Proposed solution

A parallel "Supacode Dev" app that makes the isolation assumption true by construction, rather than adding cross-process locking for a scenario production users never hit:

- Own bundle id (`app.supabit.supacode.dev`), name/icon, and `supacode-dev://` scheme → separate UserDefaults domain, separate LaunchServices/notification/deeplink identity
- State under `~/.supacode-dev` (settings, sidebar, layouts, worktree clones) → separate persisted files, and separately-restored surface UUIDs mean no shared zmx sessions
- Sparkle disabled → the dev build can't be replaced mid-session or self-update into a release build
- Socket-dir namespacing: the dev app binds its CLI socket under `/tmp/supacode-dev-<uid>` (same runtime dev-build check), each bundled CLI resolving only its own directory — so `supacode` keeps working with both apps running, and each build's CLI always reaches its own app. No separate CLI product needed: the CLI is bundled (each app ships its matching copy) and in-terminal routing is already pinned by `SUPACODE_SOCKET_PATH`.
- Make targets for the loop: `make run-app-dev`, and `make dev` (watch, rebuild, relaunch)

The prod target stays byte-for-byte unchanged. A working implementation exists in draft PR #510 (opened before the contribution policy landed — I'll bring it into compliance and rebase if approved); happy to rework the approach based on direction here.

### Alternatives considered

- Backing up / restoring `~/.supacode` by hand: error-prone, and doesn't address zmx session sharing, deeplink routing, defaults-domain bleed, or auto-update.
- Real cross-process coordination (flock/NSFileCoordinator, socket arbitration, zmx client counting): a much larger, riskier change benefiting only a scenario end users don't hit — isolation makes the whole class unreachable instead.
- An env-var override for the data directory only: fixes file clobbering but not the shared bundle id (defaults domain, deeplinks, notifications, Sparkle, LaunchServices).
- A build *configuration* instead of a dedicated target: tried in #510 — breaks `supacode-cli` (Tuist only generates SPM projects with Debug/Release) and a `PRODUCT_NAME` override collides the shared frameworks.

### Supacode version

0.10.4

### Are you planning to build this yourself?

- [x] I intend to open a pull request once this is approved and marked `ready`.

### Before submitting

- [x] I understand feature pull requests are closed until the issue is marked `ready`.

jeremybower:
Closes #635

## Summary

Dogfooding an in-development build today means running it against your real `~/.supacode` — the dev build and the daily-driver install share the bundle id, data directory, zmx sessions, deeplink routing, Sparkle updater, and CLI socket directory, so the safest workflow is "quit the app you use all day and hope nothing corrupts" (see #635 for the full failure-mode audit).

This adds **Supacode Dev**, an isolated copy that can run alongside a real install:

- Own bundle id (`app.supabit.supacode.dev`), name/icon "Supacode Dev", and `supacode-dev://` URL scheme — separate UserDefaults domain and LaunchServices/notification/deeplink identity.
- All state under `~/.supacode-dev` (settings, sidebar, layouts, worktree clones), so persisted files and restored zmx sessions never collide with prod.
- Sparkle disabled for the dev build, so it can't self-update or be replaced mid-session.
- **Socket-dir namespacing:** the dev app binds its CLI socket under `/tmp/supacode-dev-<uid>` instead of sharing `/tmp/supacode-<uid>`. The bundled CLI (shared binary, no separate product) derives dev-ness at runtime from its own executable path — the bundle id of the enclosing `.app` — resolves only its own build's socket directory, and cold-launches the app bundle it is embedded in. Each build's `supacode` always reaches its own app, and running both apps no longer breaks the CLI with "Multiple Supacode instances found".
- Make targets for the loop: `make run-app-dev` (build + launch), `make build-app-dev`, `make dev` (watch, rebuild, relaunch via watchexec), `make log-stream-dev`.

Implementation notes:

- The dev app is a **dedicated Xcode target** (`supacode-dev`) on the standard `Debug` configuration, produced by the same `supacodeAppTarget` factory as prod. A custom build configuration breaks the `supacode-cli` target (Tuist generates external SPM projects with only `Debug`/`Release`), and a command-line `PRODUCT_NAME` override collides the shared frameworks — both were tried. Frameworks, GhosttyKit, and the CLI stay single shared targets, so only the app sources compile twice, and the prod target is byte-for-byte unchanged (`Info.plist` reads name/icon/scheme via `$(VAR:default=…)`).
- Dev behaviour is decided at **runtime** from the `.dev` bundle id suffix (`SupacodePaths.isDevelopmentBuild`), not a compile flag, which would force the broken custom configuration.
- The dev app registers `supacode-dev://` for OS routing but accepts both schemes, so the CLI's `supacode://` payloads work against either build.

## Type of change

- [ ] Documentation
- [ ] Other (please describe)

## How was this tested?

Ran the full local gates on this branch after rebasing onto `main`, plus new unit tests for the CLI socket-directory resolution (`SocketDiscoveryTests`: enclosing-bundle detection, dev/prod directory choice, bare-binary and symlinked-CLI fallbacks). Verified manually that the dev app and a prod install run side by side with independent state and that each build's CLI reaches its own app.

- [x] `make test` passes

## Checklist


sbertix:
I don't know about this one, tbh 😅 do we really need this?

jeremybower:
Fair question 😄 — the itch this scratches: testing anything that touches persisted state (settings migrations, sidebar buckets, layout snapshots, worktree clone flows) means risking your real `~/.supacode` and your actual worktrees. Today the only safe option is back up the dir / use a throwaway machine. The dev build gives you a parallel install with isolated state and no auto-update, so you can break things freely and `rm -rf ~/.supacode-dev` to reset.

There's also a stability angle: I daily-drive Supacode for agent orchestration, and I've noticed running terminal sessions get flaky when two instances of the same app are open (haven't debugged it yet). Keeping in-flight changes in a separate app means the prod build I actually rely on to do the work stays untouched. `make dev` (watch + rebuild + relaunch) is the day-to-day payoff — tight edit loop without clobbering the prod app I keep open.

On cost, I tried to keep it cheap and contained:
- **Prod target is unchanged** — `Info.plist` just reads `$(VAR:default=…)`, so with no overrides it's byte-for-byte the same Supacode.
- It's one extra Xcode target sharing the same sources; frameworks/GhosttyKit/CLI are built once, so it's not a second full build.
- Dev-vs-prod is a runtime bundle-id check, no compile flags or `#if` scattered around.

That said, if it's not worth the maintenance surface to you, I'm happy to trim. Options:
1. Keep as-is.
2. Drop the separate icon/scheme polish, keep just isolated `~/.supacode-dev` state + `make dev`.
3. Park it as a doc'd local-only recipe and close the PR.

Whatever you prefer — no attachment here.

jeremybower:
@sbertix to bring this PR into compliance with the policies, I opened an issue for this feature and added more details on why it is necessary.

sbertix:
A variation of this, if not exactly this, is definitely coming. 
Same as #604, the "stalling" is not about the code implementation but the direction/abstraction itself. 

That said, to align expectations further: although I can confidently say this is eventually being merged, please do not expect this to be prioritized any further at this point ("it's not gonna happen today").

PR #511, opened 2026-06-28
Add a customizable notification sound picker

### LINKED ISSUE #464: Feature request: Customizable notification sound (sound picker)
Supacode currently plays a single, hardcoded notification sound. There's no way to choose which sound plays. I'd like a setting to pick the notification sound, possibly from (/System/Library/Sounds/) or ideally the ability to point at a custom sound file.

I don't mind doing the work and submitting a PR, if contributions are considered? 

## Current behavior

There are two separate sound paths, and neither is selectable:

1. **System notifications ON** → `SystemNotificationClient.send` builds a `UNMutableNotificationContent` and hardcodes:
   ```swift
   content.sound = .default
   ```
   (`SupacodeSettingsShared/Clients/Notifications/SystemNotificationClient.swift:133`)

2. **System notifications OFF** → an in-app `NSSound` plays the bundled `notification.wav`
   (`supacode/Clients/Notifications/NotificationSoundClient.swift`)

The Notifications settings pane only exposes a **"Play notification sound"** on/off toggle — there's no sound selector anywhere in the UI. macOS also doesn't allow changing a push notification's sound per-app, so there's no OS-level workaround either.

## Proposed behavior

Add a sound picker to **Settings → Notifications**, e.g.:

- A dropdown to choose the notification sound (System Default + a few bundled options).
- Optionally, "Choose custom file…" to point at a user-supplied `.aiff` / `.caf` / `.wav` (< 30s, per `UNNotificationSound` constraints).
- The chosen sound should apply to **both** notification paths:
  - System notifications: `content.sound = UNNotificationSound(named: UNNotificationSoundName("<chosen>.aiff"))`
  - In-app: swap the `NSSound` source accordingly.

## Why

my wife asked me to change the sound

llbbl:
Implements #464 — a sound picker in Settings → Notifications so the notification sound is selectable instead of hardcoded.

## What changed
- New `NotificationSound` enum (Shared): **System Default**, **Supacode Chime** (the bundled `notification.wav`), and the standard macOS system sounds (Funk, Glass, Ping, …).
- A `Picker` in Settings → Notifications, persisted in `GlobalSettings` with a forward/back-compatible decoder (old settings files still load; an unknown value falls back to the default).
- Applied to **both** notification paths:
  - In-app (`NotificationSoundClient`) — plays the chosen sound via `NSSound`; all options work.
  - System banner (`SystemNotificationClient`) — sets `content.sound` from the choice.

## Behavior note (intentional)
`UNNotificationSound(named:)` only resolves sounds from the app bundle / app container `Library/Sounds`, not `/System/Library/Sounds`. So on the **macOS notification banner**, only **System Default** and the bundled **Supacode Chime** carry through; the other system sounds fall back to the default banner sound. They work fully for **in-app** notifications (the default mode, where `systemNotificationsEnabled` is off). The picker caption/tooltip states this explicitly. Bundling those sounds (or copying into `Library/Sounds`) to lift the limitation could be a follow-up.

## Tests
- The selected sound is threaded through both client paths.
- `GlobalSettings` round-trips the value and defaults to System Default when the key is absent (back-compat).
- `NotificationSound` mapping unit tests (per-case banner sound, display names, system-sound names).

Closes #464


sbertix (approved):
Thanks 🙇‍♂️ 

Given the limitations of the system notification sounds, I took the liberty of moving the selector so it’s for app notifications only and updated its default.

PR #513, opened 2026-06-28
Fix supacode build warnings

jeremybower:
Fixes the compiler warnings originating in supacode's own source (dependencies and project-config items left out of scope). Each fix is its own commit.

## Changes
1. **`GhosttySurfaceView`** — `wrapperCStrings` is only read, so it's now a `let` (was `var never mutated`).
2. **`RepositoriesFeature`** — `withLock`'s returned (and intentionally ignored) `removeAnywhere` result is now `_ =`'d at both call sites (`result of withLock is unused`).
3. **`GithubSettingsView`** — dropped a dead `?? "…"` fallback; `GithubCLIError` is `LocalizedError`, so `localizedDescription` is non-optional and already carries the gateway-timeout message.
4. **Worktree views** — migrated 4 deprecated `Text + Text` concatenations to the macOS 26 string-interpolation form, preserving the inline `.foregroundStyle` styling.

## Out of scope
- `ld: ignoring duplicate libraries: '-lc++'` and the 4× "run script every build" notes live in `Project.swift` and are deliberate; changing them risks build correctness. Left as-is.

## Verification
- `make build-app` — succeeds; all 6 source warnings gone (only the out-of-scope `-lc++` linker note remains).
- `make check` — swift-format + swiftlint clean.

### DIFF
--- a/supacode/Features/Repositories/Reducer/RepositoriesFeature.swift
+++ b/supacode/Features/Repositories/Reducer/RepositoriesFeature.swift
@@ -1247,7 +1247,7 @@ struct RepositoriesFeature {
           state.resetRowLifecycleSyncBeforeReconcile(itemID: worktreeID)
           // Drop the worktree from every bucket in its section. The worktree is
           // going away entirely so its current bucket doesn't matter.
-          state.$sidebar.withLock { sidebar in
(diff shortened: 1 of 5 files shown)


sbertix (approved):
Thanks 🙇‍♂️

Issue #101, opened 2026-02-20
Worktree base directory collisions when repositories share basename


khoi:
## Summary
Supacode’s default worktree base-directory identity is not unique across repositories. It derives from repository basename, so different repos with the same folder name can collide.

## User-visible problems
- Two different repos can map to the same `~/.supacode/repos/<name>` base directory.
- Worktrees from unrelated repos can be mixed under the same parent.
- Worktree creation/listing/removal can become ambiguous or behave unexpectedly.

## Reproduction
1. Have two different git repositories with the same last path component, e.g.:
   - `/Users/me/src/org-a/app`
   - `/Users/me/src/org-b/app`
2. Add both repositories to Supacode.
3. Create worktrees in each repository.
4. Observe both repos targeting the same Supacode base path under `~/.supacode/repos/app`.

## Expected
- Different repository roots should always map to different Supacode worktree base directories by default.
- No manual configuration should be required to avoid collisions.

## Actual
- Default naming is basename-based, so repos that share a basename collide.
- Current manual `repositoryName` override can work around this, but default behavior remains collision-prone.

## Why this matters
This breaks common worktree layouts (including bare-repo + `.bare` setups) and makes repository identity fragile in real-world monorepo/multi-org directory structures.

## Suggested direction
Use a deterministic unique default directory identity derived from repository root (or remote identity), while keeping a separate user-facing display name.

asabaylus:
Separate from the silent collision: this also has a **crash**. If one repo is registered under both its working dir and its `.bare/` dir, both enumerate the same worktree → duplicate `WorktreeID` → launch crash-loops here:

```
WorktreeInfoWatcherManager.swift:116
- Dictionary(uniqueKeysWithValues: worktrees.map { ($0.id, $0) })
+ Dictionary(worktrees.map { ($0.id, $0) }, uniquingKeysWith: { first, _ in first })
```

Same guard you already use at `RepositoriesFeature.swift:4985`.

Repro: add both the repo and its `.bare/` to `repositoryRoots` with ≥1 worktree, relaunch.

<sub>Co-authored with Claude.</sub>

mvanhorn:
Implemented in https://github.com/supabitapp/supacode/pull/517 - switched the worktree lookup in setWorktrees to Dictionary(_:uniquingKeysWith:) so a duplicate WorktreeID keeps the first entry instead of trapping on launch.

Issue #226, opened 2026-04-09
Feature Request: Workspace / Context Isolation for Repositories


GuzzoLM:
### Problem

When working with multiple projects simultaneously, Supacode currently stores all repositories under a single flat structure:

```
~/.supacode/repos/<repo-name>
```

This creates a few challenges:

* No clear separation between different contexts (e.g. work, open source, side projects)
* Agents may miss project-specific context (e.g. `agents.md`)
* Harder to manage large numbers of repositories
* Increased cognitive load when switching between domains

---

### Proposal

Introduce a **Workspace (or Context) concept**, where repositories and agent context are grouped under a shared environment.

Example structure:

```
~/.supacode/workspaces/<workspace-name>/
  repos/
  agents.md
  context.json
```

---

### What a Workspace Would Represent

A workspace represents an **execution context**, including:

* A set of repositories
* Shared `agents.md` (or similar config)
* Optional context/state (e.g. tasks, environment assumptions)
* Logical grouping by purpose (e.g. `work`, `oss`, `personal`)

---

#### 1. Work vs Personal Separation

* `myCompany` workspace → company repos, strict conventions
* `myAmazingLibs` workspace → some OSS project
* `myBrandNewThing` workspace → some side project

Each with its own:

* agent behavior
* assumptions
* context

---

#### 2. Improved Agent Performance

By scoping context per workspace:

* Reduces irrelevant context passed to agents
* Avoids cross-project contamination
* Improves reasoning accuracy

---

#### 3. Better Developer Experience

* Cleaner mental model
* Easier context switching
* Aligns with how developers actually work (project-based thinking)

---

### Happy to Contribute

If this direction makes sense, I’d be happy to help iterate on the design or contribute to an implementation.

sbertix:
Hey 👋 

I definitely think this could be very useful. Only note is that I would possibly market this more as a "Profile" selection, like browsers do, since it implies much more than just a different workspace from what I can gather. 

My concern is that this would require refactoring a lot of code. Could you share a more detailed PRD and some specs on how you would want this tackled? Things like migration (cause I don't see it working as just an extra layer, tbh, but I might be wrong), #101 and #188 + `env` (it'd be nice to have them fit here), etc.

GuzzoLM:
Hi @sbertix! Thanks for the response!

On "Profile" vs "Workspace" framing, I agree. "Profile" makes more sense.

Migration path:
[...]

PR #517, opened 2026-06-29
Prevent duplicate WorktreeID crash from base-directory collisions

mvanhorn:
## Summary

Supacode no longer crash-loops on launch when a repository is registered under both its working directory and its `.bare/` directory. That layout enumerates the same worktree twice, producing a duplicate `WorktreeID`, which previously trapped while building the worktree lookup.

## Why this matters

In `WorktreeInfoWatcherManager.setWorktrees`, the lookup was built with `Dictionary(uniqueKeysWithValues:)` at `WorktreeInfoWatcherManager.swift:120`. That initializer traps at runtime on duplicate keys, so the duplicate `WorktreeID` reported in #101 crashed the app every launch. The same guard pattern (`uniquingKeysWith:`) is already used in `RepositoriesFeature.swift`.

This switches to `Dictionary(_:uniquingKeysWith:)`, keeping the first entry on collision instead of trapping. The rest of `setWorktrees` now iterates the de-duplicated values, so a duplicate ID no longer configures the same watcher or emits its immediate `filesChanged` refresh twice.

## Testing

Added `buildsWorktreeLookupWithoutTrappingOnDuplicateID` in `WorktreeInfoWatcherManagerTests.swift`: it sets two worktrees sharing one `WorktreeID` and asserts the manager initializes and emits a single `filesChanged` event. Ran swiftlint in strict mode against both changed files (0 violations).

### DIFF
--- a/supacode/Features/Repositories/BusinessLogic/WorktreeInfoWatcherManager.swift
+++ b/supacode/Features/Repositories/BusinessLogic/WorktreeInfoWatcherManager.swift
@@ -113,10 +113,8 @@ final class WorktreeInfoWatcherManager {
 
   private func setWorktrees(_ worktrees: [Worktree]) {
     let isInitialWorktreeLoad = !hasCompletedInitialWorktreeLoad && self.worktrees.isEmpty && !worktrees.isEmpty
-    // A repository registered under both its working dir and its `.bare/` dir can
-    // enumerate the same worktree twice, yielding a duplicate `WorktreeID`. Use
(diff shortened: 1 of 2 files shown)


sbertix (approved):
Thanks 🙇‍♂️ I optimized the code a bit and removed some redundant comments.  
Also, I removed the auto-close for #101, since the issue itself was actually about a much broader concern, which requires a product decision, not just a fix, and it will come together with #226.

PR #522, opened 2026-06-29
Add ⌘-click to expand or collapse all sidebar groups

VlatkoMilisav:
Adds ⌘-click on any sidebar disclosure (a repo section header or a nested branch group) to expand or collapse every group at once - the clicked item's resulting state just propagates to the whole sidebar. A new `setAllSidebarGroupsExpanded` action drives both collapse axes (repo sections and the branch-nest prefixes) and there are a couple of TestStore tests for it.

Heads up: I couldn't build locally (no Xcode 26.3 / ghostty toolchain on this machine), so it's unverified at runtime. The repo-header path leans on the native `Section(isExpanded:)` binding reading the modifier at click time, so that one's worth a quick check on device; the branch-group headers are plain Buttons so they're solid.


sbertix:
Hey 👋 thanks, just to align expectations, an "expand/collapse all" + hotkeys can definitely be added to the menu bar, etc. But a `⌘` + click is not very likely to be merged @VlatkoMilisav

VlatkoMilisav:
Makes sense - just pushed View-menu **Expand All** / **Collapse All** commands wired through `AppShortcuts` (defaults ⌘⌃] / ⌘⌃[, so they're customizable in Settings > Shortcuts like the rest). They dispatch the same `setAllSidebarGroupsExpanded` action.

I left the ⌘-click on the disclosures in as an optional extra, but happy to drop those two call sites if you'd rather keep it menu-only - just say the word.

sbertix:
Happy either way, but I'd still lean toward dropping the two `⌘`-click call sites: it's a non-standard gesture, and it isn't meaningfully fewer interactions than the menu item/hotkey already give you. It'd also retire the one path that needs on-device verification (the repo-header `Section(isExpanded:)` binding reading the modifier flags). I trust you though: if you think it's worth it, we can keep it :muscle:

Otherwise all good. We might trim a couple of comments, and the tests could go a bit further (e.g., `setAllSidebarGroupsExpandedCollapsesEveryRepositorySection` seeds no branch prefixes, so it isn't actually proving `collapsedBranchPrefixes` survive a collapse, and nothing exercises buckets beyond `.unpinned`), but it's good and a genuinely welcome addition. Thank you 🙇‍♂️

sbertix (approved):
Thanks 🙇‍♂️

Issue #391, opened 2026-06-10
Deep-link and CLI skills


blacroix:
Hi 👋🏻 

From a worktree, when I am prompting something like:

```bash
use supacode cli and deep-link to create another workspace based on this branch to plan XX
```

I expect new worktree to be created and deep-link to be used to begin the session as asked.

But actually I have to specify a bit and goes way back to understand how to do it properly:

<img width="2506" height="733" alt="Image" src="https://github.com/user-attachments/assets/5f169eb4-80a3-476b-8754-9ca94c58e199" />

Also sometime deep-link isn't properly escaped it uses `"` instead of `'`: "plan "something"" instead of 'plan "something"'

Can we make that workflow smoother?
Thanks!

sbertix:
Hey 👋 yeah, deeplinks are not part of the skills we ship. We should definitely add one skill for those too. 

If you're inside Supacode though, using the skill + the CLI already works (the agents know how to use it, if you've installed the agent integration in Settings > Developer, and the CLI is linked by default inside Supacode).

sbertix:
This is just a random test, to show it in action. 

https://github.com/user-attachments/assets/8d76c7ae-df58-47c1-b8c6-eb17de9a2338

blacroix:
Yes, the worktree creation part works; at the contrary the deep-link part (that opens a popup with the command to the targeted worktree) doesn't work very well (it needs deep-dive from Claude and sometime another prompt). I will try the 'run XX part' though. I keep you posted!

PR #531, opened 2026-06-30
docs: document supacode:// deep links in CLI agent skill

mvanhorn:
## Summary

The installed CLI agent skill (`CLISkillContent`) documents the `supacode` command but says nothing about the `supacode://` URL scheme, even though `DeeplinkParser` already handles a rich set of routes. This adds a "Deep links" section to the shipped skill content so coding agents know the scheme exists and which routes are valid.

## Why this matters

In #391 a user described prompting an agent to "use supacode cli and deep-link to create another workspace" and watching it stall: the agent had to reverse-engineer the URL shape because nothing in the shipped skill describes it, and it sometimes produced badly quoted links. The CLI commands are documented but the deep-link surface that `DeeplinkParser` parses is invisible to the agent, so each link becomes guesswork. Documenting the scheme alongside the existing command reference gives agents the same first-class coverage for deep links that they already have for the `supacode` subcommands, which is exactly the "add one skill for those too" direction a maintainer suggested on the issue.

## Changes

- Added a shared `deepLinksSection` string and interpolated it into every shipped agent variant (`claudeSkill`, `codexSkillMd`, `codexAgentsMd`, `kiroSkillMd`, `kimiSkillMd`, `piSkillMd`, `opencodeSkillMd`, and the `copilotSkillMd` alias). A single source keeps the section in parity across variants instead of eight copies that can drift.
- The section documents the `supacode://<host>/<path>[?query]` shape, percent-encoding of worktree/repo IDs and query values (the escaping pain point from the issue), and the full route list sourced from `DeeplinkParser`: worktree select/run/stop/archive/unarchive/delete/pin/unpin, per-script run/stop, tab and surface create/focus/split/destroy, `repo/open`, `repo/<id>/worktree/new`, and the `settings` routes.
- Added `CLISkillContentTests` asserting every shipped variant contains the scheme and the Deep links heading, so a future variant added without the section fails the test rather than silently shipping without deep-link coverage.

No runtime behavior changes; `DeeplinkClient` is the reference for the routes and is not modified.

### DIFF
--- a/SupacodeSettingsShared/BusinessLogic/CLISkillContent.swift
+++ b/SupacodeSettingsShared/BusinessLogic/CLISkillContent.swift
@@ -7,6 +7,38 @@ nonisolated enum CLISkillContent {
     + " Use when running Supacode CLI commands, managing worktrees, tabs, and surfaces programmatically,"
     + " or when inside a Supacode terminal session."
 
+  private static let deepLinksSection = """
+    ## Deep links
+
+    The URL scheme is `supacode://<host>/<path...>[?query]`. Worktree and repo IDs are percent-encoded
+    paths, so encode `/` as `%2F`. Query values must be URL-encoded; for example, an `input` command
+    containing spaces or quotes must be percent-encoded.
+
+    - `supacode://` - bring Supacode to front.
+    - `supacode://help` - open help.
+    - `supacode://worktree/<id>` - select/focus worktree.
+    - `supacode://worktree/<id>/run` - run the worktree's primary script.
+    - `supacode://worktree/<id>/stop` - stop scripts.
+    - `supacode://worktree/<id>/archive` - archive worktree.
+    - `supacode://worktree/<id>/unarchive` - unarchive worktree.
+    - `supacode://worktree/<id>/delete` - delete worktree.
+    - `supacode://worktree/<id>/pin` - pin worktree.
+    - `supacode://worktree/<id>/unpin` - unpin worktree.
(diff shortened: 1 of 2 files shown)


sbertix (changes requested):
Thank you 🙇‍♂️ 

I do have a concern with this, though, which is also why the phrasing of the prompt you're quoting (`use supacode cli and deeplink to create another workspace`) is inherently problematic: we don't want agents with immediate and direct access to the CLI (and can/should do everything with the CLI) to weigh deeplinks at the same level. 

That said, provide them with some knowledge of deeplinks too is definitely a welcome addition but:
1. it should be its own separate skill (the existing one is called `supacode-cli`, so, for symmetry, we can have a new `supacode-deeplinks` one)
2. it should have a "disclaimer" warning the agent to always prefer `supacode-cli`, since they usually have terminal access, and the CLI allows for chaining commands, etc, **unless**:
    1. The skill is called directly
    2. It's explicitly specified and singled-out ("use deeplinks" ✔️; "do it using cli and deeplinks" = "I'm not sure how it works but just do it" ❌)
    3. The environment does not support the command line or `supacode` is not linked/available

Could you apply the changes, please? 🙇‍♂️ 

(Also, I'm removing the auto-closing of the original issue, because that would possibly boil down to adding a single CLI command/deeplink to create a new worktree and pass a command directly to it without threading completions, etc.)

sbertix:
Closing as stale.

PR #535, opened 2026-06-30
Find GitHub CLI in common install paths

### LINKED ISSUE #509: gh cli not found
I install gh cli via homebrew. Still could not detect my gh cli install.

Image

Image

Image

wolfyy970:
## Summary

Fixes #509.

Supacode now falls back to common GitHub CLI install locations when `which gh` cannot find `gh` from the app process environment.

This covers common Homebrew and user-local installs:

- `/opt/homebrew/bin/gh`
- `/usr/local/bin/gh`
- `$HOME/.local/bin/gh`

## Why

When Supacode runs as a macOS app, its environment may not include the same `PATH` a user has in their interactive shell. That can make a valid Homebrew `gh` install appear unavailable, causing the GitHub settings screen to show that GitHub CLI is not found.

## Validation

- Added a regression test for fallback executable resolution when both normal and login-shell `which gh` probes fail.
- Ran focused `GithubCLIClientTests`.
- Ran `make build-app`.
- Manually verified the Debug app detects `gh` and shows the signed-in GitHub account.

### DIFF
--- a/supacode/Clients/Github/GithubCLIClient.swift
+++ b/supacode/Clients/Github/GithubCLIClient.swift
@@ -251,14 +251,14 @@ private struct GithubPullRequestsRequest: Sendable {
   let repo: String
 }
 
-private actor GithubCLIExecutableResolver {
+actor GithubCLIExecutableResolver {
+  nonisolated private static let logger = SupaLogger("GithubCLI")
+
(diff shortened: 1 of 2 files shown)


sbertix (approved):
Thanks 🙇‍♂️ just added some negative tests, but very solid overall 💪

Issue #500, opened 2026-06-26
[Feature Request] In-app browser


stavfx:
Would be nice to have an [in-app browser like cmux](https://github.com/manaflow-ai/cmux/blob/main/docs/assets/built-in-browser.png)

anlek:
What's the benefit for having an in-app browser? I dislike having to re-sign in in in-app browsers and missing extensions also limits it a lot. I don't mind opening it in my system browser.

stavfx:
for web development agent work. 
Every worktree can have its own browser instance that the agent can control and verify its work.

anlek:
Ahh, I see. Thanks for explaining! (I'm not on contributor, I just was interested why you'd want this feature). Hopefully supabitapp team can get back to you on it.

jeremybower:
@sbertix I'd be interested in contributing a PR for this. I've done some planning around extension support -- especially for credential vaults like 1Password. Let me know if that's welcome and I can share what I have here. Of course, there are some UI questions to clarify.

jeremybower:
# Feature request: Embedded browser surfaces as peers to terminal surfaces

## Summary

Supacode's tab/split engine is hardcoded to terminals: every tab is a `SplitTree<GhosttySurfaceView>`, and there is no notion of "surface content kind" anywhere. This issue proposes making the split-tree leaf content-agnostic so a non-terminal surface (a `WKWebView`-backed browser) can slot into the *existing* split/tab machinery, then adding an embedded browser surface on top of that seam.

The key framing: a browser is a **surface**, a peer to a terminal surface — a leaf in the split tree. That single model gives you both options at once: a browser alone in a tab reads as "a browser tab," and a browser split beside a terminal lets you reference a live page while you work. No parallel system, no choosing between the two.

The motivating use case is the loop Supacode is built around: an agent spins up a dev server in a worktree, and you want to *see* the result next to the agent without leaving the app — a browser pane split beside the terminal, pointed at `localhost:<port>` for that worktree. A later, optional follow-on is letting the agent itself drive that browser (navigate / snapshot / click) over the existing CLI/socket.

I'd suggest landing this as **two PRs**: a pure refactor that introduces the content seam (no behavior change), then a feature PR that adds the browser surface. Splitting it this way keeps the risky structural change isolated from — and reviewable independently of — the new feature.

## Motivation

[...]

PR #536, opened 2026-06-30
Make the split-tree leaf content-agnostic and rebuild pane drag-and-drop above it

jeremybower:
Groundwork for #500. Prepares the terminal split tree to host non-terminal surfaces (e.g. a browser) by making the leaf content-agnostic, and rebuilds pane-rearrange drag-and-drop as a layer that sits *above* the leaves so it works for any surface kind. No browser code is included here.

### What's in this PR

**1. Content-agnostic split-tree leaf**
- Introduces a `SurfaceView` base (the leaf *is* the real content view — first responder, drag source, AX node) with a `SurfaceContent` enum; `GhosttySurfaceView` becomes one concrete kind. The tree is typed `SplitTree<SurfaceView>` throughout, and kind-specific code routes through `content` so adding a new leaf kind breaks the build at every site that must handle it rather than failing a silent downcast.

**2. Surface focus-claim lifecycle lifted into the base**
- Moves the re-attachment focus-reclaim logic (SwiftUI detaches/reattaches sibling surfaces during split rebuilds; AppKit doesn't auto-promote a reattached view) onto `SurfaceView`, parameterized by overridable hooks, so every surface kind gets correct focus behavior without terminal-specific code.

**3. All-AppKit pane-rearrange drag layer**
- A kind-agnostic drag layer above the leaves: `SurfaceDragCoordinator` (in-flight state + drop routing), `SurfaceDragHandleView` (`NSDraggingSource` — OS drag image + reliable begin/end signal), and `SurfaceDropCatcherView` (a real AppKit drop target). Concrete `SurfaceView`s carry no rearrange code.
- The catcher is mounted per pane for the pane's lifetime and registered solely for a private surface drag type, so it's ready before a drag session starts (AppKit can miss a target that registers mid-drag), wins the drag even over a greedy child view, and stays click-through until a drag is in flight. A content drop (a Finder file) carries a disjoint type and falls through to the surface, so the terminal's own file/text drop is unchanged.

### Testing
- Unit tests for the drag coordinator lifecycle, payload parsing, and the tree-mutation drop path; existing terminal-state/split-tree tests updated for the content-agnostic leaf.
- `make build-app`, `make check`, and the test suite pass. The live AppKit drag was verified manually in the dev app.

Relates to #500.

### DIFF
--- a/supacode/Features/Terminal/Models/SurfaceDragCoordinator.swift
+++ b/supacode/Features/Terminal/Models/SurfaceDragCoordinator.swift
@@ -1,0 +1,60 @@
+import AppKit
+import SupacodeSettingsShared
+import UniformTypeIdentifiers
+
+private let dragLogger = SupaLogger("SurfaceDrag")
+
+/// Coordinates pane-rearrange drag-and-drop for one worktree's terminal area.
+///
+/// The rearrange machinery (drag handle + drop catcher) lives in a layer above
+/// the split-tree leaves, so concrete `SurfaceView`s stay agnostic to it. This
+/// object is the single piece of state that layer needs:
+///
+/// - The drag *source* (`SurfaceDragHandleView`) flips `draggingSourceID` via
+///   `beginDrag`/`endDrag` from its `NSDraggingSource` callbacks. The view layer
+///   observes `isDragging` to mount a `SurfaceDropCatcherView` over every visible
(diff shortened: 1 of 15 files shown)


jeremybower:
Will open a new PR as discussed in #500.

PR #553, opened 2026-07-02
Add visible pane title badges for split terminals

ygpark80:
## Summary
- show a small title badge on each split terminal pane
- use the surface title, then the working directory basename, then a short surface id as fallback
- keep the badge non-interactive so pane focus and drag/drop behavior are unchanged

## Testing
- xcrun swiftc -parse supacode/Features/Terminal/Views/TerminalSplitTreeView.swift

Note: make check could not run locally because mise/tuist are not installed in this environment.


ygpark80:
Closing this for now.

PR #556, opened 2026-07-02
Completion-based CLI socket acks with a configurable timeout

### LINKED ISSUE #529: supacode tab new and surface split always time out the ack (~5s, exit 1) but still create the surface
## Summary

`supacode tab new` and `supacode surface split` always fail their acknowledgement with `Error: Timed out waiting for response from Supacode` after a fixed ~5 second wait, exit 1, and print nothing to stdout. The tab or surface is actually created; only the ack times out. Because stdout is empty, the new resource UUID the command is supposed to print is lost, which breaks any script that captures the id to target the resource afterward.

## Environment

- Supacode 0.10.4 (build 1782346525)
- macOS 26.5.1 (25F80)

## Steps to reproduce

```sh
WT=$(supacode worktree list | head -1)
supacode tab new -w "$WT" -i "echo hi"; echo "exit=$?"
```

The command hangs ~5s, prints `Error: Timed out waiting for response from Supacode`, reports `exit=1`, and prints no UUID. `supacode tab list -w "$WT"` afterward shows the new tab is present and the seeded command ran. `surface split` behaves the same way.

## Expected

`tab new` / `surface split` print the new resource UUID to stdout and exit 0 once the surface is created.

## Actual

Exit 1, ~5s wall time, stderr `Error: Timed out waiting for response from Supacode`, empty stdout. The surface is created regardless.

## What I tested

I ran each command against a local worktree and captured the exit code, wall time, stdout, and whether the action took effect (by diffing `tab list` / `surface list` / `worktree list` before and after).

Always times out (exit 1, ~5s, surface still created):

- `tab new`
- `surface split`

The ~5s is constant and independent of the seeded `-i` command. Bare shell (no `-i`), `-i "echo hi"`, and `-i "sleep 600"` all timed out at 5.1 to 5.3 seconds, so it is the surface-spawn acknowledgement that never returns, not the seeded command.

Clean (exit 0, under 0.4s, correct result):

- reads: `worktree list`, `tab list`, `surface list`, `repo list`, `socket`
- `tab focus`, `tab close`, `surface close`
- `worktree focus`, `worktree pin`, `worktree unpin`, `worktree archive`, `worktree unarchive`
- `repo worktree-new`
- `worktree delete`

So the timeout is specific to the two commands that spawn a terminal surface, not to mutations or resource creation in general. `repo worktree-new` creates a worktree and acks fine.

One intermittent case: `worktree unarchive` against a worktree under `~/.supacode/repos/` timed out the same way on two occasions but was clean on later retries. I could not pin a deterministic trigger for that one.

## Impact

The CLI guidance is to capture the printed UUID from `tab new` / `surface split` to target the new resource. That is impossible here, since stdout is empty on the timeout. A caller also cannot use the exit code to tell success from failure, because the command exits 1 even though it worked. The workaround is to ignore the error and recover the new id out of band by diffing `tab list` / `surface list` before and after, which also makes blind retries dangerous: a retry spawns a duplicate tab or surface.

Possibly related: #391 (same deeplink-to-spawn path).

sbertix:
Closes #529.

`supacode tab new` and `supacode surface split` (and the other socket commands) failed their acknowledgement with `Error: Timed out waiting for response from Supacode` after ~5s and exited 1, even though the tab/surface was created. This reworks the CLI control protocol so a command holds its socket response open until the operation is observably complete, bounded by a configurable timeout, and prints the created resource id.

## What changed

**Transport (`AgentHookSocketServer`)** — the root cause: responses were not reliably delivered to a client whose socket was inherited by a forked shell. The accepted and listening fds now get `FD_CLOEXEC`, the client fd gets `SO_NOSIGPIPE`, and responses are written then half-closed with `shutdown(SHUT_WR)` before `close`, so EOF is delivered regardless of inherited dups. Responses can carry a created-resource `id`.

**App (completion acks)** — a command's socket response is deferred until the operation is observably complete, bounded by a per-fd, generation-stamped timeout watchdog (so a recycled fd can't be drained by a stale watchdog):
- `tab new` / `surface split` resolve when the surface exists, returning its id.
- `worktree-new` resolves on the new worktree's first tab and returns the worktree id, for every local path: explicit `--branch`, random name, and the interactive creation prompt (the pending id threads through the prompt and is drained if the prompt is cancelled, dismissed, or superseded). Remote creation stays immediate-ack (see follow-ups).
- `tab close` / `surface close` / git worktree delete / folder delete resolve when the resource is gone.
- Synchronous / no-signal commands ack immediately.
- Tab/surface spawn failures surface as an explicit error instead of a generic timeout; duplicate explicit ids are rejected; a delete targeting an already-terminating worktree is rejected up front; surface-close acks are scoped by worktree; socket-backed confirmation dialogs get their own bounded, token-guarded watchdog.

**CLI (`supacode-cli`)** — a `--timeout` option (default 180s, `0` = wait indefinitely), propagated to the socket read window and embedded in the deeplink for the app-side watchdog. Creation commands print the created id (`tab new`, `surface split`, `repo worktree-new`). The list underline is only emitted to a TTY, so a captured / piped id stays clean.

## Tests

New coverage in `AppFeatureCommandAckTests` for the ack state machine (resolution, timeout drain, stale-token guard, worktree-new bind/fail/cancel, duplicate-id rejection, terminating-delete guard, confirmation timeout + stale-token), plus updates to the affected deeplink tests. `make test` green.

## Follow-ups
- Remote `worktree-new` returns immediate-ack with no id; the ssh create + reload path has no first-tab signal and needs its own completion event.
- Confirmation-gated socket commands give the confirmation phase and the execution phase separate timeout budgets, while the CLI read window is a single `timeout + grace`; a late confirm followed by a slow operation could time the CLI out first. Worth a shared confirm+execute deadline.

### DIFF
--- a/supacode-cli/Commands/OpenCommand.swift
+++ b/supacode-cli/Commands/OpenCommand.swift
@@ -6,7 +6,9 @@ struct OpenCommand: ParsableCommand {
     abstract: "Bring Supacode to the front."
   )
 
+  @OptionGroup var timeoutOption: TimeoutOption
+
   func run() throws {
-    try Dispatcher.dispatch(deeplinkURL: DeeplinkURLBuilder.open())
+    try Dispatcher.dispatch(deeplinkURL: DeeplinkURLBuilder.open(), timeoutSeconds: timeoutOption.timeout)
   }
 }
--- a/supacode-cli/Commands/RepoCommand.swift
+++ b/supacode-cli/Commands/RepoCommand.swift
@@ -19,8 +19,10 @@ extension RepoCommand {
   struct List: ParsableCommand {
     static let configuration = CommandConfiguration(abstract: "List repositories.")
 
+    @OptionGroup var timeoutOption: TimeoutOption
+
     func run() throws {
-      let items = try QueryDispatcher.query(resource: "repos")
+      let items = try QueryDispatcher.query(resource: "repos", timeoutSeconds: timeoutOption.timeout)
       for item in items {
         print(item["id"] ?? "")
       }
@@ -33,8 +35,13 @@ extension RepoCommand {
     @Argument(help: "Absolute path to the repository.")
     var path: String
 
+    @OptionGroup var timeoutOption: TimeoutOption
+
     func run() throws {
-      try Dispatcher.dispatch(deeplinkURL: DeeplinkURLBuilder.repoOpen(path: path))
+      try Dispatcher.dispatch(
+        deeplinkURL: DeeplinkURLBuilder.repoOpen(path: path),
+        timeoutSeconds: timeoutOption.timeout
+      )
     }
   }
 
@@ -62,14 +69,18 @@ extension RepoCommand {
     @Option(help: "Parent directory the worktree folder is created in.")
     var location: String?
 
+    @OptionGroup var timeoutOption: TimeoutOption
+
     func run() throws {
       let rID = try resolveRepoID(repo)
-      try Dispatcher.dispatch(
+      let id = try Dispatcher.dispatch(
         deeplinkURL: DeeplinkURLBuilder.repoWorktreeNew(
           repoID: rID,
           options: .init(branch: branch, base: base, fetch: fetch, name: name, location: location)
-        )
+        ),
+        timeoutSeconds: timeoutOption.timeout
       )
+      if let id { print(id) }
     }
   }
 }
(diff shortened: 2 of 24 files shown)


mtomov:
This change fixed a pain I've experienced! Thank you for this and supacode, it's rarely to see so much thought put into a tool, it give you all the right amount of primitives, and combined with the CLI, I can automate so many flows! Thank you!

PR #562, opened 2026-07-02
fix(notifications): mute system notification for the viewed surface

### LINKED ISSUE #558: System notification fires even when the notifying surface is focused and the app is active
## What happens

With **Settings → Notifications → System Notifications** enabled and the macOS alert style for Supacode set to **Persistent** (System Settings → Notifications → Supacode), every agent notification (hook-sourced OSC 3008 `kind=notify`, or the agent's own OSC 9) lands as an alert that stays on screen until manually dismissed — even when the user is actively looking at the very surface that produced it: Supacode frontmost, worktree selected, surface focused. Watching an agent work means dismissing a pile of alerts about the session you are already reading.

Persistent is the natural choice for these notifications — a temporary banner is easy to miss when you step away, which is their whole point. The same applies to the fallback notification sound.

## Why

- In `AppFeature`, the `.terminalEvent(.notificationReceived(...))` handler gates the `systemNotificationClient.send(...)` effect only on `settings.systemNotificationsEnabled` — there is no check against the selected worktree, focused surface, or key window.
- `SystemNotificationClient`'s `willPresent` returns `[.badge, .sound, .banner]` unconditionally, so macOS's default "suppress banners for the foreground app" behavior never applies. (That override is needed for the "app active but different worktree selected" case — which means focus gating has to happen upstream, where focus is actually known.)
- The terminal layer already computes almost exactly the right signal: `WorktreeTerminalState.appendNotification` derives `isRead = isSelected() && isFocusedSurface(surfaceID)` — but it only marks the in-app inbox entry as read and is not forwarded with the event.

## Expected

A banner/sound fires only when the user is not already looking at the surface: window not key/visible, another worktree selected, or another tab/surface focused — the same composite the inbox already treats as "read on arrival" (plus the window-key check it is currently missing).

## Suggested fix

Extend the `isRead` composite with the window state the state object already tracks (`lastWindowIsKey` / `lastWindowIsVisible` from `syncFocus`), pass the flag through `onNotificationReceived` → `TerminalClient.Event.notificationReceived`, and gate the `systemNotificationClient.send` / `notificationSoundClient.play` effects in `AppFeature` on it.

Happy to submit a PR.

mykhaliuk:
Fixes #558

## What

- `WorktreeTerminalState.appendNotification` now derives `isViewed`: the worktree is selected, the surface is the focused one of the selected tab, and the window is key and visible (`lastWindowIsKey` / `lastWindowIsVisible` from `syncFocus`; `isSelected()` guards against stale window flags after switching worktrees). The same composite replaces the previous `isRead` derivation, which was missing the window-key check.
- The flag rides along `onNotificationReceived` → `TerminalClient.Event.notificationReceived(isViewed:)`, and `AppFeature` skips the `systemNotificationClient.send` / `notificationSoundClient.play` effects when the surface is viewed.
- As discussed in the issue, the behavior is exposed as a **Mute notifications for active surface** toggle under _Settings → Notifications_ (`GlobalSettings.muteNotificationsForActiveSurface`, default `true`), applied symmetrically to the banner and the fallback sound. The toggle disables itself when both delivery channels are off, driven by `SettingsFeature.State.hasActiveNotificationChannel`.

## Untouched on purpose

- The sidebar `worktreeNotificationReceived` action still fires for every notification.
- `SystemNotificationClient.willPresent` keeps forcing banners for the active app: a notification from a non-selected worktree must still show while Supacode is frontmost.

## Tests

- `AppFeatureSystemNotificationTests`: viewed + mute-on skips the banner and the sound; viewed + mute-off fires both; existing cases updated for the new event arity.
- `SettingsFeatureTests.hasActiveNotificationChannelReflectsDeliveryToggles`: all four delivery-toggle combinations.
- `AgentBusyStateTests`: callback closures updated.

### DIFF
--- a/SupacodeSettingsFeature/Reducer/SettingsFeature.swift
+++ b/SupacodeSettingsFeature/Reducer/SettingsFeature.swift
@@ -49,6 +49,7 @@ public struct SettingsFeature {
     public var inAppNotificationsEnabled: Bool
     public var notificationSoundEnabled: Bool
     public var systemNotificationsEnabled: Bool
+    public var muteNotificationsForActiveSurface: Bool
     public var moveNotifiedWorktreeToTop: Bool
     public var analyticsEnabled: Bool
     public var crashReportsEnabled: Bool
@@ -81,6 +82,12 @@ public struct SettingsFeature {
     public var repositorySettings: RepositorySettingsFeature.State?
     @Presents public var alert: AlertState<Alert>?
 
+    /// True when at least one notification delivery channel (macOS banner or
+    /// the fallback sound) can fire, so surface-mute has something to mute.
+    public var hasActiveNotificationChannel: Bool {
+      systemNotificationsEnabled || notificationSoundEnabled
+    }
+
     public init(settings: GlobalSettings = .default) {
       let normalizedDefaultEditorID = OpenWorktreeAction.normalizedDefaultEditorID(settings.defaultEditorID)
       appearanceMode = settings.appearanceMode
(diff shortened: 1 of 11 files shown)


sbertix (approved):
Thanks 🙇‍♂️

PR #563, opened 2026-07-02
Add JetBrains Rider support

DaniBedz:
**JetBrains Rider integration:**

- Added JetBrains Rider as per the current pattern.
- Verified as working via local build

Thanks in advance.


sbertix (approved):
Thank you 🙇‍♂️

PR #575, opened 2026-07-03
Add a worktree color CLI command and deeplink

iannotian:
## Motivation

The app already supports a per-worktree sidebar tint (the worktree customization sheet sets a `RepositoryColor`, persisted in `sidebar.json`), but there's no programmatic way to set it. External orchestrators want one — e.g. a ticket poller that colors each worktree by workflow state: blue while an agent is working, red when a reviewer requests changes, green when it's out for review. `pin`/`archive`/`delete` are already scriptable; this gives `color` the same parity.

## What this does

```
supacode worktree color <value> [-w <id>]
```

`<value>` is one of the seven predefined names (`red|orange|yellow|green|teal|blue|purple`), `#RRGGBB[AA]` hex, or `none` to clear the override. It dispatches `supacode://worktree/<id>/color?value=<v>` through the exact same pipeline as `pin`/`unpin`: CLI → `DeeplinkURLBuilder` → `DeeplinkClient` → `AppFeature` → `RepositoriesFeature`.

image

## Design notes

- The write goes through a new color-only `SidebarState.setColor` primitive, so an existing **title override is preserved** (`setCustomization` overwrites both fields; there's an end-to-end test asserting preservation).
- The CLI target only links ArgumentParser, so it validates with a small local allowlist (ASCII-hex only) and fails fast with the accepted values; the app re-validates authoritatively via `RepositoryColor.parse`. An unrecognized value at the app layer surfaces the standard alert, so the socket ack correctly reports failure to the CLI.
- Docs updated: CLI reference view, deeplink reference view, and all agent CLI-skill content variants.

## Tests

- `DeeplinkClientTests`: predefined name, percent-encoded hex, `none`, missing/empty value, trailing-slash ID normalization.
- `AppFeatureDeeplinkTests`: routing incl. title preservation end-to-end, hex → `.custom`, `none` clears color only, invalid value alerts without touching sidebar state.
- `SidebarStateTests`: `setColor` title preservation, nil-clears, unpinned fallback, `Item` wire-format round-trip.

Heads-up on verification: the CLI target was compiled and exercised directly (deeplink output verified byte-exact) and swift-format/swiftlint pass clean, but I couldn't run the app test suites locally — happy to iterate on anything CI flags.

## Out of scope (possible follow-ups)

- Exposing color in `supacode worktree list` output.
- A `worktree title` command over the same seam.


sbertix (changes requested):
Hey 👋 thanks.

Could we structure this so that title and color are not separate units, like in the app, and, most importantly, they're readable too? For example, `supacode worktree appearance -w <w_id>` to read it, and `supacode worktree appearance -w <w_id> --title … (optional) --color (optional)` (or with its own verb, idk) to update it.

iannotian:
@sbertix addressed your concerns! Thank you for the feedback.

sbertix (approved):
Thanks 🙇‍♂️

iannotian:
Thank you @sbertix. Excellent software.

PR #579, opened 2026-07-05
Sidebar surfaces ⌃ worktree shortcut hints when ⌘ is held

slgoodrich:
## Scenario

Sidebar visible with worktree rows. User holds **⌘**.

## Expected

No shortcut hints on the worktree rows. Worktree selection binds to **⌃1..⌃0** (`AppShortcuts.selectWorktree*`, `modifiers: [.control]`), so ⌘ never triggers those shortcuts — nothing to hint.

## Observed

⌃-based hints appear on every numbered worktree row while ⌘ is held. `CommandKeyObserver.shouldShowShortcuts` returned `true` for ⌘ **or** ⌃, and `SidebarListView` gated its hint dictionary on that single union flag.

Symmetric second bug from the same union: holding **⌃** lit the **⌘**-based tab-bar hints (`TerminalTabView`, tab selection is ⌘1..⌘9).

## Root cause

One boolean (`isPressed`) tracked "⌘ or ⌃ held" but was consumed by two surfaces with opposite modifiers — the sidebar (⌃) and the tab bar (⌘). Either modifier lit both surfaces.

## Fix

Split `CommandKeyObserver` into independent `isCommandPressed` / `isControlPressed` flags:

- Sidebar worktree hints → `isControlPressed`
- Tab-bar hints → `isCommandPressed`

Tests updated to cover each modifier independently.

### DIFF
--- a/supacode/App/CommandKeyObserver.swift
+++ b/supacode/App/CommandKeyObserver.swift
@@ -1,17 +1,23 @@
 import AppKit
 import SwiftUI
 
-/// Tracks whether the user is currently holding ⌘ or ⌃ so the UI can surface shortcut hints.
+/// Tracks whether the user is currently holding ⌘ or ⌃ so the UI can surface
+/// shortcut hints. The two modifiers are tracked independently: the tab bar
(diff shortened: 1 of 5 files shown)


sbertix:
This is intended and symmetric on purpose.

PR #586, opened 2026-07-06
Fix non latin layout

ERuban:
## Summary

Fix terminal control shortcuts on non-Latin keyboard layouts.

Supacode’s Ghostty surface bridge was deriving control-key text from the active keyboard layout. With a non-Latin layout selected, a physical `Ctrl-C` could arrive as a Cyrillic/non-ASCII codepoint instead of ASCII `c`, so terminal programs such as Claude would not receive the expected interrupt.

This change makes terminal `Ctrl` chords layout-independent by resolving the physical keyCode through the existing US-QWERTY shortcut fallback before passing the unshifted codepoint to Ghostty.

## Details

  - Applies only to terminal `Ctrl` chords without `Cmd` or `Option`.
  - Keeps regular text input layout-aware.
  - Keeps app/menu shortcuts untouched.
  - Covers common terminal control keys beyond letters, including punctuation and space,
  e.g. `Ctrl-/`, `Ctrl-[`, `Ctrl-\`, `Ctrl-Space`.


ERuban:
Have no chance to test it, bcs I'm on newest macos and xcode beta versions.
I'd appreciate it if someone could test it.

Issue #419, opened 2026-06-19
Recursively discover and nest worktrees under a folder


piotrkowalczuk:
I routinely work across several monorepos at once. I use [zordon](https://github.com/piotrkowalczuk/zordon) to manage *compound workspaces*, where each workspace is a single folder holding multiple sparse-checkout worktrees — often spanning more than one monorepo. An agent then operates on a subset of those worktrees under the task directory.

What I'm missing is a bird's-eye view. When a folder contains nested git worktrees, I'd like Supacode to discover them recursively and display them nested under that folder, each with its own status. That alone would give me an at-a-glance read of every unit of work in the task and where each one stands.

To be clear, this isn't about capability — as @sbertix noted on #226, an agent isn't bound to the worktree it's launched from and can already juggle sibling directories. The gap is visibility: today the sidebar is organized `Repo → Worktree`, so the worktrees that make up one compound workspace scatter under their respective repos. There's no single place that shows "this folder = these N worktrees" and their combined state.

```text
A compound workspace on disk (laid out by zordon):

  ~/work/TICKET-1234/                # a compound workspace = one folder
  ├── Alphasfile                 # zordon: one isolated stack for the whole folder
  ├── src/payments/                  # sparse worktree of monorepo A (subset of dirs)
  ├── src/orders/                    # sparse worktree of monorepo A (different subset)
  └── src/ledger/                    # sparse worktree of monorepo B
        ↳ an agent operates on a subset of these under the folder

What I'd like in the sidebar (recursive discovery under the folder):

  TICKET-1234                        # the folder, surfaced as a unit of work
  ├── payments   ● dirty   ↑2   PR #41 open
  ├── orders     ✓ clean        running: dev
  └── ledger     ✓ clean   ↑1   PR #44 draft
```

The behavior I have in mind is purely reflective: Supacode doesn't create or own these worktrees, it just walks the folder, finds the nested git worktrees (across any number of monorepos), and groups them under that folder node. Per-worktree status would carry the usual signals — branch, ahead/behind, dirty, PR, running script, agent state — and a folder-level rollup would summarize them. It reuses the worktree awareness Supacode already has; it's keyed by the containing folder instead of by repo.

Why it helps: any compound layout — hand-made, scripted, or produced by an orchestrator like [zordon](https://github.com/piotrkowalczuk/zordon) (which already stands up one isolated stack per workspace folder) — gets a task-level cockpit for free, with no new orchestration in Supacode. Just discovery and a nested view.

[...]

PR #588, opened 2026-07-06
Show a folder's git repositories nested beneath its sidebar row

jtfile-wq:
## What

Folders added to the sidebar now show the git repositories inside them, attached beneath the folder row:

- Each first-level child repo renders as **one compact row** (its main checkout) with the same info a top-level repo shows — branch, PR state, check status.
- A connector rail and indentation make the attachment visually obvious; a chevron on the folder row collapses/expands the group.
- **Folder behavior is unchanged**: clicking the folder still opens the folder-root terminal session.
- Clicking an attached repo row stays **inside the folder's session, scoped to that repo**: it focuses an existing tab/pane whose shell is working in that repo, or opens a new tab in the folder session rooted at the repo directory. Top-level repos keep their existing behavior.

## Why

A common workflow is running a coding agent from a folder root that contains several related repos (e.g. `unify/` holding `unify-api` + `unify-app`) so the agent sees all of them in one session. Today that costs you all per-repo visibility in the sidebar. This keeps the folder-root workflow intact while restoring branch/PR/checks status per repo.

## Implementation notes

- Child repos are **derived at load time** (one-level scan of the folder), never persisted to `repositoryRoots` — they appear/disappear with the filesystem.
- Linked worktree checkouts (`.git` file pointing into `/worktrees/`) are skipped so agent worktrees don't show up as repos; submodules still surface.
- Removing a folder cascades to its derived children; children that fail to load still render (with the failure), matching top-level behavior.
- Parent/child matching is trailing-slash-normalized on both sides (open-panel URLs carry a trailing slash, round-tripped path ids don't).
- Folder rows with attached children are excluded from pinned/active hoisting so children are never orphaned from their parent.
- Scoped click routing resolves a pane's directory via live pwd (OSC 7), then the zmx session registry's `start_dir` (zmx doesn't forward OSC 7), then the surface's launch directory.

## A design question for maintainers

If a repo is both explicitly added at top level *and* discovered as a folder child, it renders once, nested. We chose to hide "Remove Repository…" on nested rows instead of tracking provenance, because removal would be silently undone by re-discovery on the next load regardless. If you'd prefer provenance tracking (explicit adds stay top-level), happy to rework.

## Tests

~15 new tests covering discovery (worktree-checkout skip, missing dirs), ordering, slash-divergent parent matching, section structure, collapse, removal cascade, failure-row visibility, and selection redirect; plus a zmx list-parser test for `start_dir`. Full suite passes.

### DIFF
--- a/supacode/Clients/Zmx/ZmxClient.swift
+++ b/supacode/Clients/Zmx/ZmxClient.swift
@@ -273,6 +273,11 @@ nonisolated enum ZmxSessionListParser {
     var name: String
     /// nil when the count is unknown (err/status line); the reaper spares these.
     var clients: Int?
+    /// Directory the session's shell was spawned in (`start_dir=`). The
+    /// folder-scoped tab matcher uses it to identify which pane works in
+    /// which repo, since zmx doesn't forward the shell's OSC 7 pwd reports.
+    /// Defaulted so the reaper's name/clients call sites stay source-stable.
+    var startDir: String? = nil
   }
 
   static func parse(_ stdout: String) -> [Entry] {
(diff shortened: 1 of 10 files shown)


sbertix:
This was already addressed several times (e.g., #419), and it’s not going to be added, even if the current PR implemented a proper UI (which it currently does not 😢). 

We can 100% add creation-level quick-detect and add repos from a given folder, as mentioned before, but there’s no current plan to modify sidebar nesting/group at this point.

PR #604, opened 2026-07-08
Adopt externally managed zmx sessions

### LINKED ISSUE #591: Adopt externally managed remote terminal sessions into Supacode
### What problem does this solve?

I use Supacode as the local command center, but some coding-agent sessions are created and managed outside Supacode on a remote development host. Those sessions already run under zmx/tmux-like persistence and have stable working directories, repository roots, branch names, titles, and agent identity.

Today Supacode can manage remote SSH repositories and persistent sessions it creates itself, but there is no supported way for an external orchestrator to tell Supacode: “this remote session exists; show it in the sidebar, attach a tab to it, preserve a stable identity, and update/remove it idempotently.”

The workaround is to repeatedly drive Supacode through private/local automation. That works, but repeated no-op updates can cause visible sidebar churn, especially when several external sessions map to the same remote repo/folder.

### Proposed solution

Add a small, documented adoption/sync API for externally managed sessions.

A possible shape:

- A CLI command such as `supacode session adopt --manifest <file>` or `supacode bridge apply --file <file>`
  - Manifest fields:
  - external namespace + session id
  - remote host alias
  - repository root
  - worktree/working directory
  - optional branch/title/agent
  - attach command, e.g. `ssh -tt host zmx attach <session>`
  - stable tab id or external id used to derive one
- The operation should be idempotent:
  - unchanged manifests should not rewrite app state or visibly reorder/sidebar-refresh rows
  - stable IDs should preserve tabs across repeated syncs
  - missing sessions may be pruned explicitly or after a threshold

A minimal first version could avoid owning the external lifecycle: Supacode would only adopt/focus/reattach sessions that already exist, not create or delete them.

### Alternatives considered

- Use remote SSH repositories directly in Supacode: works for sessions Supacode creates, but does not adopt sessions created by another orchestrator.
- Use deep links / existing CLI commands per tab: possible, but there is no stable “ensure this external session exists exactly once” primitive.
- Keep a local bridge script: works as a prototype, but relies on private behavior and can cause UI churn if it has to repeatedly apply the same state.

### Supacode version

_No response_

### Are you planning to build this yourself?

- [x] I intend to open a pull request once this is approved and marked `ready`.

### Before submitting

- [x] I understand feature pull requests are closed until the issue is marked `ready`.

LucasIcarus:
Closes #591

## Summary

Adds a `supacode tab adopt-zmx <session>` command (and a matching `…/tab/adopt-zmx` deeplink) so an external orchestrator can surface a zmx session it already manages as a Supacode tab. Supacode only attaches — it never creates or kills the external session. Re-running is idempotent: the tab id is derived from the session name, an unchanged sync neither steals focus nor churns the sidebar, and closing or pruning an adopted tab spares the external session (only Supacode-owned sessions are torn down). Adopted tabs are excluded from the layout snapshot, so a restart defers re-adoption to the owner instead of restoring a detached shell.

## Type of change

- [ ] Documentation
- [ ] Other (please describe)

## How was this tested?

- Unit tests cover the deeplink parse/build, idempotent re-adoption without focus stealing, snapshot exclusion of adopted tabs, and that tab close / prune / surface close never kill an externally owned session.
- `make check` and `make test` pass locally, and I built, ran, and dogfooded the app.

Dogfooding used a small external orchestrator (a `launchd` agent) that syncs live zmx sessions from a remote devbox into Supacode. Simplified model:

```python
# launchd agent, polls the devbox every few seconds
NS = uuid.uuid5(uuid.NAMESPACE_URL, "supacode-bridge")
stable_tab_id = lambda host, sid: str(uuid.uuid5(NS, f"{host}/{sid}"))

def sync(cli, host):
    sessions = inventory_live_zmx(host)          # ssh host, cross-checked against `zmx list --short`
    digest = sha256(canonical(sessions))
    if digest == last_applied:                   # unchanged manifest -> no-op (no churn)
        return
    known = set(run(cli, ["worktree", "list"]))
    for s in sessions:                           # one worktree tab per live remote session
        wt = host + s.worktree_path              # remote worktree id
        if wt in known:
            run(cli, ["tab", "adopt-zmx", s.zmx_name,
                      "-w", wt, "--id", stable_tab_id(host, s.id), "--title", s.title])
    for gone in missing_for_n_polls():           # prune stale tabs only after a threshold
        run(cli, ["tab", "close", "-w", gone.wt, "-t", gone.stable_tab_id])
    save(digest)
```

The stable UUID pins each remote session to one tab across polls; the manifest digest makes an unchanged sync a true no-op. Running this every few seconds against real devbox sessions is what surfaced the focus-steal, sidebar-churn, and external-session-kill cases fixed here.

- [x] `make test` passes

- Model(s): Claude Opus 4.8 & GPT 5.5
- Harness / tools: Cursor & Codex

## Checklist

### DIFF
--- a/supacode-cli/Commands/TabCommand.swift
+++ b/supacode-cli/Commands/TabCommand.swift
@@ -1,4 +1,5 @@
 import ArgumentParser
+import CryptoKit
 import Foundation
 
 struct TabCommand: ParsableCommand {
@@ -9,6 +10,7 @@ struct TabCommand: ParsableCommand {
       List.self,
       Focus.self,
       New.self,
+      AdoptZmx.self,
       Close.self,
     ],
     defaultSubcommand: Focus.self
@@ -90,6 +92,71 @@ extension TabCommand {
     }
   }
 
+  struct AdoptZmx: ParsableCommand {
+    static let configuration = CommandConfiguration(
+      commandName: "adopt-zmx",
+      abstract: "Add or move an existing zmx session into a worktree tab."
+    )
+
+    @Argument(help: "Existing zmx session name.")
+    var session: String
+
+    @Option(name: [.short, .long], help: "Worktree ID. Defaults to $SUPACODE_WORKTREE_ID.")
+    var worktree: String?
+
+    @Option(help: "Tab title to show in Supacode.")
+    var title: String?
+
+    @Option(name: [.short, .customLong("id")], help: "Stable UUID for the adopted tab.")
+    var newID: String?
+
+    @OptionGroup var timeoutOption: TimeoutOption
+
+    func run() throws {
+      let sessionID = try ZmxSessionArgument.normalized(session)
+      let worktreeID = try resolveWorktreeID(worktree)
+      let resolvedID: String
+      if let newID {
+        resolvedID = try Self.validatedTabID(newID)
+      } else {
+        resolvedID = Self.stableTabID(sessionID: sessionID)
+      }
+      try Dispatcher.dispatch(
+        deeplinkURL: DeeplinkURLBuilder.tabAdoptZmx(
+          worktreeID: worktreeID,
+          sessionID: sessionID,
+          title: title,
+          id: resolvedID
+        ),
+        timeoutSeconds: timeoutOption.timeout
+      )
+      print(resolvedID)
+    }
+
+    private static func stableTabID(sessionID: String) -> String {
+      let digest = SHA256.hash(data: Data("supacode:zmx-session:\(sessionID)".utf8))
+      var bytes = Array(digest.prefix(16))
+      bytes[6] = (bytes[6] & 0x0f) | 0x50
+      bytes[8] = (bytes[8] & 0x3f) | 0x80
+      return UUID(
+        uuid: (
+          bytes[0], bytes[1], bytes[2], bytes[3],
+          bytes[4], bytes[5],
+          bytes[6], bytes[7],
+          bytes[8], bytes[9],
+          bytes[10], bytes[11], bytes[12], bytes[13], bytes[14], bytes[15],
+        )
+      ).uuidString
+    }
+
+    private static func validatedTabID(_ raw: String) throws -> String {
+      guard let uuid = UUID(uuidString: raw) else {
+        throw ValidationError("Invalid --id value: expected a UUID.")
+      }
+      return uuid.uuidString
+    }
+  }
+
   struct Close: ParsableCommand {
     static let configuration = CommandConfiguration(abstract: "Close a tab.")
 
@@ -109,5 +176,24 @@ extension TabCommand {
         timeoutSeconds: timeoutOption.timeout
       )
     }
+  }
+}
+
+private nonisolated enum ZmxSessionArgument {
+  static func normalized(_ value: String) throws -> String {
+    let trimmed = value.trimmingCharacters(in: .whitespacesAndNewlines)
+    guard !trimmed.isEmpty else {
+      throw ValidationError("zmx session name cannot be empty.")
+    }
+    guard trimmed != ".", trimmed != "..",
+      !trimmed.unicodeScalars.contains(where: isRejectedScalar)
+    else {
+      throw ValidationError("zmx session name cannot be '.', '..', contain '/', or contain control characters.")
+    }
+    return trimmed
+  }
+
+  private static func isRejectedScalar(_ scalar: Unicode.Scalar) -> Bool {
+    scalar.value == 0 || scalar.value == 0x2F || CharacterSet.controlCharacters.contains(scalar)
   }
 }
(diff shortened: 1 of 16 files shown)


LucasIcarus:
## Quirk: an adopted zmx session running a mouse-tracking TUI (Claude Code) can't scroll until the first keystroke

Hit this dogfooding `tab adopt-zmx`; posting the root cause in case there's a cleaner intended behavior.

**Symptom.** A tab adopted onto a live zmx session running Claude can't be scrolled (wheel/trackpad) — it looks frozen. Any keystroke, or a window resize, instantly fixes it. Only Claude is affected (codex/gpt don't rely on SGR mouse events).

I confirmed it's not on supacode / libghostty side: ghostty keeps `mouse_event` enabled the whole time and correctly emits the SGR wheel events to the pty — they're dropped inside zmx.

**Why.** supacode attaches as a **non-leader** client. In `Daemon.handleInput`, a non-leader's bytes only reach the pty when `util.isUserInput()` is true, and that helper intentionally excludes mouse events:

```zig
// util.zig — isUserInput()
// mouse events: CSI M / CSI < — EXCLUDE these
// only intentional keyboard input should trigger leader switch
if (csi.final == 'M' or csi.final == '<') return false;
```

```zig
// main.zig — Daemon.handleInput()
if (leader == client) { queuePtyInput(payload); return; }  // leader: forward all
if (isUserInput(payload)) {                                // non-leader: only "user input"
    try setLeader(client);                                 //   → promotes to leader
    queuePtyInput(payload);                                //   → and forwards
}
// a non-leader's mouse: neither promotes nor forwards → silently dropped
```

So the wheel events are dropped. A keystroke passes `isUserInput`, which promotes supacode to leader, after which everything (incl. the wheel) flows — hence "press a key to unfreeze".

Both intents behind this look right to me: **leader = resize authority**, and **ambient mouse motion/focus shouldn't steal the leader** from whoever's typing. The quirk is that `isUserInput` gates *both* "promote leader" *and* "forward to pty", so excluding mouse from the leader switch also drops it entirely — even though mouse input has no resize-style single-authority requirement.

LucasIcarus:
@sbertix   This PR was ready for review now, 2-days local dogfooding works fine with it's local build version and my local link-bridge launchd scripts. 

> But indeed a quirk "bug" located and mentioned above. 

Thks for your efforts. Any following up and revision suggestions are welcome.

sbertix:
The main pain points aren’t really the issues or suggestions themselves (and there are several, since we’re here, e.g., the `supa-` namespace collision kills the adopted session, leading-dash session names are parsed as options, cross-worktree moves destroy the source tab before validating the destination, cross-host tab-ID collision on the default ID, failed adopts are acknowledged as success, "Session not found" reports success, missing tests, inaccurate docstrings, etc.). But those can be fixed easily.

I held back from giving more structured feedback before because I’m worried about the direction. The core `ownedZmxSessionIDs` refactor is sound and well tested, but the primitive itself worries me more the more I think about it (again, the implementation itself is fine).

LucasIcarus:
> The main pain points aren’t really the issues or suggestions themselves (and there are several, since we’re here, e.g., the `supa-` namespace collision kills the adopted session, leading-dash session names are parsed as options, cross-worktree moves destroy the source tab before validating the destination, cross-host tab-ID collision on the default ID, failed adopts are acknowledged as success, "Session not found" reports success, missing tests, inaccurate docstrings, etc.). But those can be fixed easily.
> 
> I held back from giving more structured feedback before because I’m worried about the direction. The core `ownedZmxSessionIDs` refactor is sound and well tested, but the primitive itself worries me more the more I think about it (again, the implementation itself is fine).

@sbertix  Thanks for the detailed read — and agreed the worthy worries.   I only use supacode to serve my use case and didn't hold the critical big picture, happy to keep it out of core until/unless the shape feels right. No pressure either way.

And if I understand the situation correctly, the scroll quirk above plus the failure modes you listed are symptoms of the same thing: the primitive has supacode *display and act on a session it doesn't own*, so it inherits a large, leaky surface (lifecycle, kill-safety, ID identity, even the multiplexer's input model) it can't fully control.

If there *is* any appetite, I'd rather find the smallest contract everyone would be comfortable owning than push the current surface. A maximally narrowed version in my mind might be:

- **attach / display / detach only** — never create, kill, or prune an external session (supacode owns nothing on the other side);
- **require an explicit stable id** — no default-derived id, so no cross-host collision;
- **drop cross-worktree move entirely** (the most destructive path);
- **strict session-name validation** (reject leading-dash / the `supa-` namespace).

But still, I don't know supacode's full workflow surface or codebase well enough to judgee, and don't obtain rich variety of workflow variants to dogfooding with supacode. So I'll gladly defer to whichever direction you prefer — including shelving this.

sbertix:
I'm worried about the opposite, tbh: that the contract is not "big enough" to actually treat attached sessions as fully fledged Supacode ones. We're introducing a second-class surface, without indication, without further escalation possibilities. 
I think it'd probably make more sense to double down the other way: once you attach a session, it's 1:1 a Supacode session. But I'm not sure yet either 😅

LucasIcarus:
Understood. I'll keep running it locally. Whenever the shape feels right to you, I'm happy to reshape or redo (or some other PR) this to match.

PR #607, opened 2026-07-08
Fix Grok presence hooks by forwarding SUPACODE_* env into subprocesses

### LINKED ISSUE #603: Grok presence badges missing — hook subprocess lacks SUPACODE_* env
### What happened?

After installing Grok integration hooks from Supacode Settings (v0.10.5), Grok lifecycle hooks are written to `~/.grok/hooks/supacode.json` but presence badges and the Active sidebar never update when Grok runs in a Supacode terminal.

I expected installing hooks from Settings to make Grok show the same presence badges as Claude/Codex. Instead, hooks run silently with no OSC 3008 output — the surface never appears in Active and no Grok badge is shown.

Root cause appears to be that Grok spawns hook commands in a subprocess that does **not** inherit the Supacode terminal's `SUPACODE_*` environment. The installed commands guard on `[ -n "${SUPACODE_SURFACE_ID:-}" ]`, so the guard always fails and the hook no-ops. Manually adding per-hook `env` blocks that forward `${SUPACODE_SURFACE_ID}`, `${SUPACODE_SOCKET_PATH}`, etc. (per Grok hooks docs) restores badge behavior.

### Steps to reproduce

1. Install or reinstall Grok hooks from Supacode Settings → Integrations → Grok (tip build 0.10.5 with PR #600 integration).
2. Open a Supacode terminal surface in any repository.
3. Start a Grok session in that surface.
4. Submit a user prompt (or trigger a tool use).
5. Observe the sidebar: no Grok presence badge and the surface does not appear in Active.
6. (Optional confirmation) Inspect `~/.grok/hooks/supacode.json` — managed hooks have inline `compositeCommand` shell but no `env` blocks. Add `env` passthrough for `SUPACODE_*` vars → badges work after Grok reloads hooks.

0.10.5 (1783464897)

### macOS version

macOS 26.5.1

### System locale

en_US

### Mac hardware

M3 Pro

### Relevant logs or screenshots

Installed hook command (abbreviated) from `~/.grok/hooks/supacode.json`:

```
[ -n "${SUPACODE_SURFACE_ID:-}" ] && { __tty=$(ps -o tty= -p "$PPID" ...); printf '\033]3008;start=grok;event=busy...' > "$__tty"; } >/dev/null 2>&1 || true # supacode-managed-hook
```

No `env` block on the hook entry. With `SUPACODE_SURFACE_ID` unset in the hook subprocess, the guard short-circuits and nothing is written to the tty.

### Are you planning to fix this yourself?

- [x] I intend to open a pull request that fixes this bug.

### Before submitting

- [x] I am on the latest available version of Supacode.

ehthayer:
Closes #603

## Summary

Grok spawns hook commands in a subprocess that does not inherit the Supacode terminal's `SUPACODE_*` environment. The installed composite commands guard on `SUPACODE_SURFACE_ID`, so the guard always fails and presence badges silently no-op.

This change adds per-hook `env` blocks with `${VAR}` expansion (per Grok hooks docs) to forward all canonical `SUPACODE_*` vars into hook subprocesses. Existing installs missing these blocks are reported as `.outdated` so Settings can prompt re-install / auto-update.

## Type of change

- [ ] Documentation
- [ ] Other (please describe)

## How was this tested?

- [x] `make test` passes (2,317 tests in 156 suites)

**Unit / integration coverage**
- `GrokHookSettingsTests.everyHookForwardsSupacodeEnv` — every managed hook carries `AgentHookSettingsCommand.grokHookEnvPassthrough`
- `GrokSettingsInstallerTests.installStateReturnsOutdatedWhenEnvPassthroughMissing` — stale installs without `env` report `.outdated`
- `GrokSettingsInstallerTests.installAllHooksWritesSupacodeManagedFile` — install writes `${SUPACODE_SURFACE_ID}` env block

**Local dogfood (Supacode terminal → Grok hooks → presence)**
1. Reinstalled via `GrokSettingsInstaller.installAllHooks()` from this branch; `installState()` → `installed`.
2. Confirmed `~/.grok/hooks/supacode.json`: every `supacode-managed-hook` entry has per-hook `env` passthrough for all `SUPACODE_*` vars.
3. Reproduced #603: hook subprocess with no `SUPACODE_*` → guard short-circuits, no `OSC 3008` output.
4. Verified fix: same hook with env passthrough → `OSC 3008;start=grok;event=busy`.
5. Fired the installed hook in a live Supacode surface (`SUPACODE_SOCKET_PATH` → running app); `~/.supacode/layouts.json` updated to `agent=grok`, `activity=busy` on that surface.

`grok -p` single-turn was attempted but did not return within ~3 min (API latency); the hook → OSC → app path above covers the integration under test.

- Model(s): Grok (xAI)
- Harness / tools: Grok CLI / Cursor

## Checklist

### DIFF
--- a/SupacodeSettingsShared/BusinessLogic/AgentHookCommandOwnership.swift
+++ b/SupacodeSettingsShared/BusinessLogic/AgentHookCommandOwnership.swift
@@ -22,6 +22,12 @@ nonisolated enum AgentHookCommandOwnership {
     {
       return true
     }
+    // Early Grok presence helper installed under ~/.grok/hooks/bin/ without
+    // the ownership sentinel. Fingerprint the helper basename so install
+    // prunes dual legacy+composite hook maps left from pre-envelope Grok.
+    if command.contains(AgentHookSettingsCommand.legacyGrokOSCHelperMarker) {
+      return true
+    }
     // Pre-envelope hooks carry the verbatim 4-var presence-guard but
     // neither the sentinel nor the CLI shim. The guard is a Supacode-
     // specific fingerprint: a user following the documented single-var
--- a/SupacodeSettingsShared/BusinessLogic/AgentHookCommandOwnership.swift
+++ b/SupacodeSettingsShared/BusinessLogic/AgentHookCommandOwnership.swift
@@ -22,13 +22,6 @@ nonisolated enum AgentHookCommandOwnership {
     {
       return true
     }
-    // Early Grok presence helper under ~/.grok/hooks/bin/ without the
-    // ownership sentinel. Fingerprint the full relative path (not bare
-    // basename) so install prunes dual legacy+composite maps without
-    // claiming unrelated user commands that mention `supacode-osc.sh`.
-    if command.contains(AgentHookSettingsCommand.legacyGrokOSCHelperMarker) {
-      return true
-    }
     // Pre-envelope hooks carry the verbatim 4-var presence-guard but
     // neither the sentinel nor the CLI shim. The guard is a Supacode-
     // specific fingerprint: a user following the documented single-var
--- a/SupacodeSettingsShared/BusinessLogic/AgentHookCommandOwnership.swift
+++ b/SupacodeSettingsShared/BusinessLogic/AgentHookCommandOwnership.swift
@@ -22,9 +22,10 @@ nonisolated enum AgentHookCommandOwnership {
     {
       return true
     }
-    // Early Grok presence helper installed under ~/.grok/hooks/bin/ without
-    // the ownership sentinel. Fingerprint the helper basename so install
-    // prunes dual legacy+composite hook maps left from pre-envelope Grok.
+    // Early Grok presence helper under ~/.grok/hooks/bin/ without the
+    // ownership sentinel. Fingerprint the full relative path (not bare
+    // basename) so install prunes dual legacy+composite maps without
+    // claiming unrelated user commands that mention `supacode-osc.sh`.
     if command.contains(AgentHookSettingsCommand.legacyGrokOSCHelperMarker) {
       return true
     }
--- a/SupacodeSettingsShared/BusinessLogic/AgentHookPayloadSupport.swift
+++ b/SupacodeSettingsShared/BusinessLogic/AgentHookPayloadSupport.swift
@@ -36,9 +36,25 @@ nonisolated struct AgentCommandHook: Encodable {
   let type = "command"
   let command: String
   let timeout: Int
+  let env: [String: String]?
 
-  init(command: String, timeout: Int) {
+  init(command: String, timeout: Int, env: [String: String]? = nil) {
     self.command = command
     self.timeout = timeout
+    self.env = env
+  }
+
+  func encode(to encoder: Encoder) throws {
+    var container = encoder.container(keyedBy: CodingKeys.self)
+    try container.encode(type, forKey: .type)
+    try container.encode(command, forKey: .command)
+    try container.encode(timeout, forKey: .timeout)
+    if let env, !env.isEmpty {
+      try container.encode(env, forKey: .env)
+    }
+  }
+
+  private enum CodingKeys: String, CodingKey {
+    case type, command, timeout, env
   }
 }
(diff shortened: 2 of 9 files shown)


ehthayer:
@sbertix doing the dogfooding locally now that i skipped before!

ehthayer:
@sbertix local dogfooding details in the PR description

sbertix:
Thanks for this 🙏 I pushed a small follow-up commit on top ([`5a20342c`](https://github.com/supabitapp/supacode/commit/5a20342c1dffe1b5aa49f3cb60657ab9257ce457)):

- The `installStateReturnsOutdatedWhenEnvPassthroughMissing` test was passing for the wrong reason. Its fixture had a single hook, so the base command-set check already returned `.outdated` and `managedHooksLackEnvPassthrough` never ran. I rebuilt it from the full canonical hook map with `env` stripped so the state actually flips on the env check, and added missing-key and drifted-value cases plus `AgentCommandHook` env-omission tests.
- `managedHooksLackEnvPassthrough` now logs the read/decode failure via `SupaLogger` instead of swallowing it, and I dropped its unused `fileManager` parameter.

Since the original repro was manual, could you pull the branch and confirm the badges still light up end to end? Reinstall the Grok hooks from Settings, start a Grok session in a surface, and check the Active sidebar + presence badge update. Just want to be sure the follow-up did not disturb the behavior your change fixes before this goes in. 🙇‍♂️

ehthayer:
@sbertix re-dogfooded on top of your follow-up (`5a20342c`):

- Reinstalled Grok hooks via the PR installer path → `integrationState=installed`
- Every managed hook still has the full `SUPACODE_*` `${VAR}` env map
- Guard still short-circuits with no env (the original #603 failure mode) and passes with passthrough
- Live `grok -p` on a fresh surface flipped presence **busy → idle** in `layouts.json` (reply `pong`)

Follow-up looks good to land — badges still light up end to end.

ehthayer:
@sbertix follow-up hardening on the branch (`8dbb3e0c`) 🙏 :

## What

Harden Grok hook install-state after the env-passthrough fix: one disk read for command-set + env checks, prune leftover `supacode-osc.sh` hooks, and close test gaps so the env detector can’t pass for the wrong reason.

## Why

Review residual from the env-passthrough path: `installState` re-read the hooks file for env and fail-opened to `.installed` if that read failed; early `~/.grok/hooks/bin/supacode-osc.sh` entries lacked the ownership sentinel and survived reinstall as dual maps; `everyHookForwardsSupacodeEnv` dropped hooks missing `env` via `compactMap`.

## How

- `AgentHookSettingsFileInstaller.installState` takes `additionalOutdatedIfInstalled` and runs it on the same parsed snapshot
- `GrokHookSettings.managedHooksLackEnvPassthrough(in:)` no longer opens the file itself
- `AgentHookCommandOwnership` treats `supacode-osc.sh` as legacy-managed (prune on install)
- Tests: single-event env missing, user-authored isolation, post-install detector happy path, legacy ownership + prune, full per-hook env completeness

## Test

Unit coverage above; dogfood reinstall → `integrationState=installed`, 8 managed / 0 legacy / full env map.

## Out of scope

- Shared multi-agent model for non-command install drift (still Grok-specific env gate)
- Changing OSC emit / presence wire format

ehthayer:
@sbertix tiny follow-up on the legacy fingerprint (`a0507822`):

Review flagged matching bare `supacode-osc.sh` as too wide for shared ownership — any agent install/uninstall could strip a user hook that merely mentioned that basename. Unlikely in practice, but cheap to tighten: ownership now matches the full path shape `.grok/hooks/bin/supacode-osc.sh` (what the early Grok helper actually installed), plus a negative test that bare basename is not owned.

sbertix:
Was `~/.grok/hooks/bin/supacode-osc.sh` ever shipped? 😅 I can't seem to find it. 
It looks like you're tying to prune things that were never there on the remote, so those would need dropping.

sbertix:
Also, in general, we should only concern ourselves with migrating users when features get into the regular release channel and Grok was never supported in a non-tip release.

sbertix (approved):
Thanks 🙇‍♂️

PR #610, opened 2026-07-09
Fix Hermes presence between turns

### LINKED ISSUE #611: Hermes presence disappears between turns instead of going idle
## Problem

Hermes currently tears down its presence at the end of a response turn. The plugin emits `session_end`, which makes the app remove the badge once the turn finishes.

That makes Hermes behave differently from the other agents: they return to `idle` between turns, so the badge stays visible without indicating active work.

## Expected behavior

When a Hermes response turn completes, Hermes should emit `idle` instead of `session_end`. The badge should remain visible between turns, but show the idle state rather than looking busy.

## Related PR

Fixed by #610.

panosAthDBX:
## Summary
- Keep the Supacode-managed Hermes environment block stable across turns
- Update the Hermes installer test expectations

## Testing
- make build-app

Closes #611


panosAthDBX:
Linked issue #611 in the PR description. The missing linked-issue policy requirement is now addressed.

sbertix:
I'm not sure I understand this, and I don't really have `hermes` set up to test it.

Is this PR basically supposed to hide the badge while the agent is idle? That's what I gathered from the issue.
If so, this would make it behave differently from all other agents, and I would keep the pre-existing behavior.

panosAthDBX:
In fact, the opposite. I misrepresented the issue in the initial bug report, sorry.

The problem I'm seeing is that Hermes is currently the odd one out: at the end of a turn it emits `session_end`, which makes the app tear the badge down. Other agents go back to `idle`, so they remain present between turns without looking busy.

So this PR is not trying to hide the badge while the agent is idle. It changes Hermes to emit `idle` at the end of a turn, so the badge stays visible but no longer indicates active work. That should make Hermes behave more like the other agents, not differently from them.

I'll update #611 / the wording to make that clearer.

sbertix (approved):
Thanks 🙇‍♂️

PR #638, opened 2026-07-11
Fix coalescesBurstOfProgressReports flake by advancing until the flush lands

### LINKED ISSUE #636: coalescesBurstOfProgressReports is flaky under load — TestClock advance races the throttle task's sleep registration
### What happened?

During a full `make test` run (parallel testing, host under load), `GhosttySurfaceBridgeTests/coalescesBurstOfProgressReports` failed; re-running the class in isolation passes every time. I expected the test to be deterministic.

The race: `settleThenAdvance` (GhosttySurfaceBridgeTests.swift) does a best-effort `Task.megaYield()` before `clock.advance(by:)`. Under load, the `progressFlushTask` spawned by `scheduleProgressFlush()` (GhosttySurfaceBridge.swift) can fail to reach its `clock.sleep(for: progressThrottleInterval)` before the advance runs. The advance then finds no registered sleeper, the flush task sleeps from the already-advanced instant, and — with no further advances in this test — the trailing flush never fires: `progressValue` stays `10` and `callbackCount` stays `1`, failing the two final expectations.

The sibling test `staleProgressClearsAfterTimeout` already documents and works around this exact race ("The stale watch can register its sleep after the first advance under load, so keep advancing until the REMOVE lands") with a bounded advance-until-settled loop; `coalescesBurstOfProgressReports` has no such loop. Mirroring that pattern — advance in a bounded loop until `progressValue == 50` — would make it deterministic, matching the intent that the bound only guards a regression.

### Steps to reproduce

1. Run the full suite (`make test`) while the machine is otherwise busy — the failure is probabilistic and load-dependent.
2. Observed once in a parallel full-suite run; the class passes when run in isolation, consistent with the yield-based settling losing only under scheduler pressure.

0.10.4 (dev build from main @ b1525fab)

### macOS version

macOS 26.5

### System locale

en_CA

### Mac hardware

M2 Pro

### Relevant logs or screenshots

```shell
Test case 'GhosttySurfaceBridgeTests/coalescesBurstOfProgressReports()' failed on 'My Mac - supacode (22361)' (0.000 seconds)
```

### Are you planning to fix this yourself?

- [x] I intend to open a pull request that fixes this bug.

### Before submitting

- [x] I am on the latest available version of Supacode.

jeremybower:
Closes #636

## Summary

`coalescesBurstOfProgressReports` advanced the `TestClock` exactly once, so under load the throttle task spawned by `scheduleProgressFlush()` could register its `clock.sleep` after the advance, leaving the trailing flush unfired (`progressValue` stuck at 10, `callbackCount` at 1). Test-only fix: advance in a bounded loop until `progressValue` reaches 50, mirroring the advance-until-settled pattern the sibling `staleProgressClearsAfterTimeout` already uses; the bound only guards a regression. `callbackCount == 2` stays exact because a flush task that wakes with nothing pending exits without applying or rescheduling, so extra ticks cannot over-fire the callback.

## Type of change

- [ ] Documentation
- [ ] Other (please describe)

## How was this tested?

Ran `make check`, the `GhosttySurfaceBridgeTests` class in isolation (all 17 cases pass, including `coalescesBurstOfProgressReports`), the full `make test` suite (2381 passed), and `make build-app`. The full suite has one failure that is unrelated and pre-exists on `main`: `PullRequestMergeQueueStatusTests/surfacesPositionAndEstimatedTime` hardcodes "~10 min left" but an `en_CA` locale formats "~10 mins left"; it fails identically without this change (this diff touches only `GhosttySurfaceBridgeTests.swift`). I'll file that separately.

- [x] `make test` passes
- [ ] I built and ran the app to confirm the change works

## Checklist

### DIFF
--- a/supacodeTests/GhosttySurfaceBridgeTests.swift
+++ b/supacodeTests/GhosttySurfaceBridgeTests.swift
@@ -200,8 +200,14 @@ struct GhosttySurfaceBridgeTests {
     #expect(bridge.state.progressValue == 10)
     #expect(callbackCount == 1)
 
-    // One throttle tick flushes only the latest coalesced value.
-    await settleThenAdvance(clock, by: .milliseconds(50))
+    // One throttle tick flushes only the latest coalesced value. The flush
+    // task can register its sleep after the advance under load, so keep
+    // advancing until the trailing flush lands; the bound only guards a
+    // regression. Extra ticks can't over-fire the callback: a flush task that
+    // wakes with nothing pending exits without applying or rescheduling.
+    for _ in 0..<50 where bridge.state.progressValue != 50 {
+      await settleThenAdvance(clock, by: .milliseconds(50))
+    }
     #expect(bridge.state.progressValue == 50)
     #expect(callbackCount == 2)
   }


sbertix:
Thanks 🙇‍♂️

Issue #465, opened 2026-06-23
Branch-update PRs (merging main into feature) incorrectly shown on the default branch


geoberle:
### Description

When someone opens a PR that merges `main` into a feature branch to bring it up to date (usually done by mistake instead of rebasing), Supacode shows that PR when I'm on main locally. That PR has nothing to do with my `main` checkout, it's just updating the feature branch. Also, this is a situation you can't fix because such PRs can't be deleted retroactively.

Example: https://github.com/Azure/ARO-HCP/pull/981 merges `main` into a feature branch
The PR is merged and the target branch deleted. Supacode still shows it on my local main.

<img width="312" height="64" alt="Image" src="https://github.com/user-attachments/assets/880687b4-4de1-4b75-b29c-3afd119704c5" />

### What's happening

The PR matching in GithubGraphQLPullRequestResponse.swift filters candidates in three groups: upstream (same-repo), fork, and deleted-fork PRs. The fork and deleted-fork filters already have a `baseRefName != branch` guard that catches this, but the upstream filter doesn't, it accepts all PRs where headRepository matches without checking what the PR targets.

So a PR like this slips through:
* headRefName = main (matches my local branch)
* baseRefName = some-feature-branch
* headRepository = same repo

khoi:
Closing as a duplicate of #695, which remains open and marked ready. Both reports reproduce a merged same-repository PR whose head is the persistent default branch (main → feature/dev) appearing on the local default-branch worktree. #695 tracks that bug and PR #720 is the proposed fix. This is duplicate triage, not a claim that the behavior is fixed; the example and matcher analysis here remain useful evidence for #695.

PR #720, opened 2026-07-22
Fix stale merged PR badge on the default-branch worktree

### LINKED ISSUE #695: Default branch worktree shows a stale merged PR
### What happened?

A clean worktree checked out on the repository's **default branch** (`main`) displays a **merged** pull request badge (PR #320 "Main", merged) as if that PR belonged to the worktree.

I expected a worktree on the default branch with no pending changes to show just the branch (default), with no PR attached.

This affects any workflow that opens sync PRs *from* the default branch into another branch (e.g. `main → dev`), because those PRs have `headRefName: main` and never go away for the default branch (the branch persists after merge instead of being deleted like a feature branch).

### Steps to reproduce

1. Use a repo whose default branch is `main`, with a workflow that opens sync PRs *from* `main` into another branch (e.g. `main → dev`), so `main` has merged PRs whose `headRefName` is `main`.
2. Open the repo in Supacode with the main worktree checked out on `main`, clean, with no local changes.
3. Observe the main worktree shows a merged PR badge (the most recently updated merged `head=main` PR) instead of no PR.

### Cause (from the source)

Tracing the PR auto-matching:

- `GithubCLIClient.makeBatchPullRequestsQuery` fetches, per branch, `pullRequests(first: 5, states: [OPEN, MERGED], headRefName: "<branch>", orderBy: {field: UPDATED_AT, direction: DESC})`.
- For the default branch this matches historical `default → X` sync PRs (they use the default branch as the *head*).
- `GithubGraphQLPullRequestResponse.pullRequestsByBranch` ranks candidates by state (OPEN > MERGED) then recency, so with no open `head=main` PR it selects the newest **merged** one.
- `RepositoriesFeature.refreshRepositoryPullRequests` then assigns it to the worktree via `pullRequestsByWorktreeID[worktree.id] = prsByBranch[worktree.name]`.

The matcher never special-cases the repository's default branch, so a worktree on the default branch permanently shows a stale merged PR.

A possible direction (happy to implement once triaged): exclude the repository's default branch from PR association in `refreshRepositoryPullRequests`. One open question for triage — suppress **all** PRs on the default branch, or only merged/closed ones (keeping a currently-open sync PR visible)?

0.10.6 (1783877721)

### macOS version

macOS 26.5

### System locale

en_IL

### Mac hardware

Apple M1 (MacBookPro17,1)

---

_Disclosure: diagnosis assisted by an AI agent (Claude Code). All commits for any follow-up PR will be authored by me._

Snirpini:
Closes #695

## Summary

A worktree checked out on the repository's **default branch** showed a stale **merged** PR badge. Any PR whose `headRefName` is the default branch is a *sync* PR (e.g. `main → dev`), not the default-branch worktree's own work — and because the default branch persists after such a PR merges (unlike a feature branch, which is typically deleted), the newest merged `head=<default>` PR kept matching the default-branch worktree indefinitely.

The fix skips PR association for the worktree whose name equals the repository's default branch, in `refreshRepositoryPullRequests` — the point where PRs are mapped to worktrees. The default branch is fetched defensively with `try?`: if that lookup fails, the refresh falls back to today's behavior rather than dropping every worktree's badge on a transient `gh` hiccup.

**Scope — Option A.** This suppresses *all* PR badges on the default-branch worktree, merged or open. It's the smallest, clearest change and matches the report: a clean default-branch worktree should show just the branch, with no PR attached. A narrower **Option B** — suppress only merged/closed `head=<default>` PRs while still surfacing a currently-*open* sync PR — is also reasonable. I went with A, but the guard is localized to one spot, so switching to B (or adding it) would be a small follow-up there. Happy to change it if you'd prefer to keep open sync PRs visible on the default branch.

## Type of change

- [ ] Documentation
- [ ] Other (please describe)

## How was this tested?

Two reducer tests added in `supacodeTests/RepositoriesFeatureTests.swift`:

- `repositoryPullRequestRefreshSkipsDefaultBranchWorktree` — the default-branch worktree gets **no** PR even when `batchPullRequests` returns a merged `head=main` PR for it, while a feature worktree still gets its PR. This reproduces #695 and fails without the fix.
- `repositoryPullRequestRefreshAssignsWhenDefaultBranchLookupFails` — when the default-branch lookup throws, the refresh still assigns by branch name, guarding the defensive fallback against a future regression.

- [x] `make test` passes (run with `TEST_PARALLEL=NO`)
- [ ] I built and ran the app to confirm the change works — verified via the automated tests above; I did not run the GUI app.

- Model(s): Claude Opus 4.8
- Harness / tools: Claude Code

## Checklist

- [x] For a feature, the linked issue is labeled `ready`. (This is a bug fix; the issue is labeled `ready`.)

### DIFF
--- a/supacode/Features/Repositories/Reducer/RepositoriesFeature.swift
+++ b/supacode/Features/Repositories/Reducer/RepositoriesFeature.swift
@@ -4453,6 +4453,13 @@ struct RepositoriesFeature {
         await send(.repositoryPullRequestRefreshCompleted(repositoryID))
         return
       }
+      // A PR whose head is the repository's default branch is always a sync PR
+      // (e.g. `main -> dev`), never the default-branch worktree's own work, so it
+      // must not badge that worktree — the head=default branch persists after
+      // merge, leaving a stale merged PR permanently matched to it (#695). Fetch
+      // defensively: an unknown default branch falls back to today's behavior
+      // rather than dropping every worktree's badge on a transient `gh` failure.
+      let defaultBranch = try? await githubCLI.defaultBranch(repositoryRootURL)
       do {
         let prsByBranch = try await githubCLI.batchPullRequests(
           remoteInfo.host,
@@ -4462,6 +4469,9 @@ struct RepositoriesFeature {
         )
         var pullRequestsByWorktreeID: [Worktree.ID: GithubPullRequest?] = [:]
         for worktree in worktrees {
+          if worktree.name == defaultBranch {
+            continue
+          }
           pullRequestsByWorktreeID[worktree.id] = prsByBranch[worktree.name]
         }
         await send(
--- a/supacodeTests/RepositoriesFeatureTests.swift
+++ b/supacodeTests/RepositoriesFeatureTests.swift
@@ -6656,6 +6656,93 @@ struct RepositoriesFeatureTests {
     await store.finish()
   }
 
+  // A worktree on the repository's default branch must never show a PR badge: any
+  // PR whose head is the default branch is a sync PR (e.g. `main -> dev`), not that
+  // worktree's own work, and it persists after merge as a stale match (#695).
+  @Test func repositoryPullRequestRefreshSkipsDefaultBranchWorktree() async {
+    let repoRoot = "/tmp/repo"
+    let mainWorktree = makeWorktree(id: repoRoot, name: "main", repoRoot: repoRoot)
+    let featureWorktree = makeWorktree(
+      id: "\(repoRoot)/feature",
+      name: "feature",
+      repoRoot: repoRoot
+    )
+    let repository = makeRepository(id: repoRoot, worktrees: [mainWorktree, featureWorktree])
+    var state = makeState(repositories: [repository])
+    state.githubIntegrationAvailability = .available
+    state.reconcileSidebarForTesting()
+    // The newest merged `head=main` sync PR is what the matcher would otherwise pin to `main`.
+    let mergedMainPullRequest = makePullRequest(state: "MERGED", headRefName: "main", number: 320)
+    let featurePullRequest = makePullRequest(state: "OPEN", headRefName: "feature", number: 5)
+    let store = TestStore(initialState: state) {
+      RepositoriesFeature()
+    } withDependencies: {
+      $0.gitClient.remoteInfo = { _ in GithubRemoteInfo(host: "github.com", owner: "o", repo: "r") }
+      $0.githubCLI.defaultBranch = { _ in "main" }
+      $0.githubCLI.batchPullRequests = { _, _, _, _ in
+        ["main": mergedMainPullRequest, "feature": featurePullRequest]
+      }
+    }
+    store.exhaustivity = .off
+
+    await store.send(
+      .worktreeInfoEvent(
+        .repositoryPullRequestRefresh(
+          repositoryRootURL: URL(fileURLWithPath: repoRoot),
+          worktreeIDs: [mainWorktree.id, featureWorktree.id]
+        )
+      )
+    )
+    await store.receive(\.repositoryPullRequestRefreshCompleted)
+    await store.finish()
+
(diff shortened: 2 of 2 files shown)


Snirpini:
Rebased on `main` (now at `5b32953b`) — no conflicts, `make check` and `make test` still pass (410 tests).

Still reproduces on current `main`: `refreshRepositoryPullRequests` maps `prsByBranch[worktree.name]` with no default-branch guard, so the newest merged `head=main` sync PR stays matched to the default-branch worktree indefinitely.

On the open question in the description — I've settled on **Option A** (suppress all PR badges on the default-branch worktree). It matches the report in #695 and keeps the guard in one spot. If you'd rather keep a currently-*open* sync PR visible there, that's a small follow-up on the same line and I'm happy to do it.

No rush — just flagging that it's rebased and ready whenever it fits your queue.

sbertix:
Sorry if I'm stalling on this. A bit unsure on the best way to tackle the problem, cause this solution might not be enough.

Snirpini:
Your hesitation was justified — I think my patch fixes this in the wrong place. I've put an alternative up as a branch rather than force-pushing over your review:

**[compare: main...Snirpini:fix/sync-pr-matching](https://github.com/supabitapp/supacode/compare/main...Snirpini:supacode:fix/sync-pr-matching)**

The bottom line: instead of skipping the default-branch worktree in the reducer, drop any candidate in `pullRequestsByBranch` where `head == defaultBranch && base != defaultBranch`. A PR flowing out of the default branch into something else is a sync PR, never the head branch's own work.

That keeps matching policy in one place (next to the `baseRefName` guards from `4a15af54`), drops the `gh repo view` call my version added to every refresh, and removes the Option A/B question I left you with — as a property of the PR rather than of the worktree, there's nothing to decide. It also covers what #465 reported.

`make check` and `make test` pass; the 6 suppression tests fail with the rule disabled, so they're real regression tests.

<details>
<summary>Why the reducer was the wrong place</summary>

The `gh repo view --json defaultBranchRef` I added was uncached, so it ran per repo per refresh — a second login-shell `gh` spawn alongside `resolveRemoteInfo`. It also answers for the wrong repository: in a fork clone `gh repo view` resolves to upstream, so it returns upstream's default branch, not the one being displayed. This version resolves from local refs via `automaticWorktreeBaseRef` and caches per repository; an unresolvable default disables the rule rather than dropping badges.

</details>

<details>
<summary>Why head <em>and</em> base, and how #465 fits</summary>

Checking both keeps it principled rather than "hide things on `main`": with `dev` as the default branch, `main → dev` is a real contribution and still matches `main`, while `dev → main` is dropped from `dev`. There's a test for each direction.

On #465 — same root cause, reported earlier. One thing I couldn't make work: its suggested fix is to add the missing `baseRefName != branch` guard to the upstream tier, but as far as I can tell that condition can never hold there, since GitHub won't accept a PR whose head and base are the same ref within one repo. So the tier asymmetry looks real but harmless, and the default-branch signal is doing the actual work. Let me know if I've misread that.

I've kept `Closes #695` as the only link since #465 isn't `ready` — happy to do whatever you prefer with it.

</details>

<details>
<summary>Two gaps I left out on purpose</summary>

- A feature branch whose PR merged, wasn't deleted, and then got new commits still shows the merged badge. That needs `headRefOid` plus the local tip SHA — `git worktree list --porcelain` already parses and discards that SHA, so it's cheap, but it's a different change. Happy to open an issue.
- A branch with two open PRs picks an arbitrary winner by `updatedAt`; `baseRefName` is fetched but not used for ranking.

</details>

I don't want to assume this is the right direction — you know the sidebar's constraints far better than I do. If you'd rather keep the original, or take neither and shape it differently, that's completely fine. And if there's something I haven't covered or a concern still sitting at the back of your mind, I'd genuinely like to hear it and help get it all covered properly.

Happy to force-push into this PR, open a fresh one, or leave it — whichever suits you.

PR #739, opened 2026-07-29
Add a --background opt-out to CLI and deeplink focus-follow

### LINKED ISSUE #722: Setting (or CLI flag) to suppress focus-follow on CLI worktree/tab actions
### Problem

Every socket/CLI worktree and tab action moves the in-app selection and keyboard
focus to the affected worktree:

- `supacode repo worktree-new`
- `supacode tab new`
- `supacode worktree delete`
- `supacode worktree run` / `stop`

When Supacode is driven from external orchestration tooling — a "lead" agent
session running in one Supacode tab that programmatically creates worktrees,
spawns worker tabs, and tears them down — each of those commands yanks focus
**away from the tab I'm actively working in** and over to the newly-affected
worktree's terminal. During a busy session that's dozens of focus jumps out of
the tab I'm typing in.

To be precise about the mechanism (this is in-app focus, not window activation):
the CLI worktree-action reducer dispatches

```swift
.selectWorktree(worktreeID, focusTerminal: true)
```

`selectWorktree` sets the single-worktree selection and, with
`focusTerminal: true`, focuses that worktree's terminal — so keyboard focus
leaves the caller's tab. It's gated only by `Deeplink.WorktreeAction.selectsWorktree`,
which is hard-coded `true` for every action except `appearance` / `tabRename`:

```swift
// Deeplink.swift
/// Whether dispatching this action should also select / focus the worktree.
/// Metadata-only updates (appearance, tab rename) skip it so they don't steal focus.
var selectsWorktree: Bool {
  switch self {
  case .appearance, .tabRename: false
  default: true
  }
}
```

So the "don't steal focus" path already exists for metadata-only actions — there's
just no way to opt lifecycle actions into it. No `settings.json` key and no CLI
flag currently suppresses it.

### Request

A way to opt out of the select/focus-follow for CLI- and deeplink-issued actions,
so a programmatic driver can create/delete worktrees and spawn tabs without
stealing focus from the tab the user is in. Either:

1. **A global setting** — e.g. "Don't change focus on CLI/deeplink worktree
   actions" (analogous to the existing `automatedActionPolicy`), or
2. **A per-command flag** — `--no-focus` on `worktree-new`, `tab new`,
   `worktree delete`, `worktree run` (dispatch the action with
   `focusTerminal: false` and skip the `selectWorktree`).

Keeping the current focus-follow as the default is fine; the ask is just an
opt-out for programmatic drivers.

Today the only workaround is to snap focus back after the fact with
`supacode tab focus -w <caller-worktree> -t <caller-tab>`, which works but
produces a visible focus bounce (focus jumps to the new worktree, then back).

### Not a duplicate of #721

#721 is the app *hanging* on tab switches. This is intentional focus-follow being
unconditional for CLI actions — a separate, working-as-coded behavior I'd like to
be able to turn off.

sbertix:
Closes #722

## Summary

Every socket/CLI worktree and tab action moved the sidebar selection and
keyboard focus to the affected worktree, so a session driving Supacode
programmatically lost focus on every create and teardown.

This adds a `--background` flag (and a matching `background=true` deeplink
parameter) to every action that would otherwise focus its target:
`repo worktree-new`, `worktree run`/`stop`/`archive`/`unarchive`/`delete`/
`pin`/`unpin`, `tab new`/`close`, and `surface split`/`close`. It suppresses
both the cross-worktree selection and the intra-worktree activation: new
tabs, splits, and script tabs land in the background, and closes do not pull
focus back. The intent survives the interactive worktree-creation prompt,
the input-confirmation dialog, and the archive/delete lifecycle scripts.

Omitting the flag reproduces the previous behavior exactly, and the
dedicated `focus` commands do not accept it. Tab creation now expresses
selection and keyboard focus as one `TabActivation` value; background tabs
append so a run of them keeps its order, and the very first tab still
selects.

Also fixes bare `worktree run`/`stop` resolving their target from the
selected worktree instead of the one the command names, which suppressing
the selection would otherwise have turned into running the wrong script.

Known limitation: closing the focused surface of the currently selected tab
still transfers focus to the surviving pane, preserving the invariant that a
tab with visible leaves keeps a focused surface.

## Type of change

- [ ] Documentation
- [ ] Other (please describe)

## How was this tested?

Reducer tests cover the deeplink gate in both directions, the confirmation
round-trip, the pending-worktree path, malformed `background` values, the
tab-selection axis (ordering, first-tab selection), and the corrected
`run`/`stop` targeting against a non-selected worktree. Verified manually by
driving a second worktree from a live session with `--background` on tab
creation, splits, scripts, and lifecycle actions.

- [x] `make test` passes

## Checklist

### DIFF
--- a/SupacodeSettingsShared/BusinessLogic/CLISkillContent.swift
+++ b/SupacodeSettingsShared/BusinessLogic/CLISkillContent.swift
@@ -96,14 +96,14 @@ nonisolated enum CLISkillContent {
     supacode worktree list [-f] [--status <status>] [--not-archived] [--with-status]  # List worktree IDs (-f = focused only).
     supacode worktree status [-w <id>]                  # Read status/archived/focused for one worktree.
     supacode worktree focus [-w <id>]                   # Focus worktree.
-    supacode worktree run [-w <id>] [-c <uuid>]         # Run script (default: primary run-kind; -c for a specific UUID).
-    supacode worktree stop [-w <id>] [-c <uuid>]        # Stop script (default: all run-kind; -c for a specific UUID).
+    supacode worktree run [-w <id>] [-c <uuid>] [--background]         # Run script (default: primary run-kind; -c for a specific UUID).
+    supacode worktree stop [-w <id>] [-c <uuid>] [--background]        # Stop script (default: all run-kind; -c for a specific UUID).
     supacode worktree script list [-w <id>]             # List configured scripts (id / kind / name). Running rows are underlined.
-    supacode worktree archive [-w <id>]                 # Archive worktree.
-    supacode worktree unarchive [-w <id>]               # Unarchive worktree.
-    supacode worktree delete [-w <id>]                  # Delete worktree.
-    supacode worktree pin [-w <id>]                     # Pin worktree.
-    supacode worktree unpin [-w <id>]                   # Unpin worktree.
+    supacode worktree archive [-w <id>] [--background]                 # Archive worktree.
+    supacode worktree unarchive [-w <id>] [--background]               # Unarchive worktree.
+    supacode worktree delete [-w <id>] [--background]                  # Delete worktree.
+    supacode worktree pin [-w <id>] [--background]                     # Pin worktree.
+    supacode worktree unpin [-w <id>] [--background]                   # Unpin worktree.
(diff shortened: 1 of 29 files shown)


mtomov:
Huge, thank you! ❤️

PR #756, opened 2026-08-01
Fix remote SSH terminal compatibility

bmargula:
Closes #754

## Summary

Add a remote-side compatibility prelude for Supacode-managed interactive SSH
surfaces. When the local terminal advertises `xterm-ghostty`, the prelude checks
whether the remote host has that terminfo entry and falls back to
`xterm-256color` only when it is unavailable.

Supacode launches SSH directly instead of passing through Ghostty's normal SSH
wrapper, so fresh remote accounts currently miss that compatibility behavior.
Terminal-aware commands such as `clear`, `less`, and `tput` can consequently
fail with `unknown terminal type: xterm-ghostty`.

The check runs under POSIX `/bin/sh` before the configured remote login shell
loads its profile. Hosts with `xterm-ghostty` support retain its richer
capabilities, other TERM values remain unchanged, and non-interactive
`SSHCommand.invocation` calls are unaffected. Both interactive command shapes
share one compatibility seam.

## Type of change

- [x] Bug fix (the linked issue is a bug report)
- [ ] Feature (the linked issue is a feature request marked `ready`)
- [ ] Documentation
- [ ] Other (please describe)

## How was this tested?

- [ ] `make check` passes (format + lint)
- [ ] `make test` passes
- [ ] I built and ran the app to confirm the change works

Completed locally:

- The mise-pinned `swift-format lint --strict` passes for both changed files.
- `swiftc -frontend -parse` accepts both changed files.
- `git diff --check` passes.
- Shell behavior tests cover supported `xterm-ghostty`, missing/unknown
  `xterm-ghostty` (including an unavailable `infocmp`), and an unrelated TERM.
- A real SSH PTY preserves `xterm-ghostty` when the remote entry exists and
  resolves to `xterm-256color` when a missing entry is simulated; `tput clear`
  exits successfully in both cases.

The repository commands were attempted but cannot complete on this machine:

- `make check` completes formatting, then SwiftLint cannot load SourceKit
  because no full Xcode toolchain is installed.
- `make test` stops in the repository preflight because Xcode 26.3 is absent.
- `make doctor` otherwise passes and reports Xcode 26.3 as its only remaining
  prerequisite.

The three test/build checkboxes above should remain unchecked until those
commands pass with Xcode 26.3 and the app has been exercised from a built copy.

## AI tool disclosure (optional)

- Model(s): GPT-5
- Harness / tools: Codex

## Checklist

- [x] This pull request is linked to an issue with `Closes #` above.
[...]

PR #767, opened 2026-08-04
Fix remote shell quoting for fish login shells

### LINKED ISSUE #585: Remote surfaces fail when the host login shell is fish (nested quoting unparseable)
Every new tab on a remote worktree exits immediately, showing either "Process exited" or a wall of mangled `'\''\'\''` text. Regression from #574.

Image

## Root cause

sshd parses the remote command with the user's login shell — fish on the affected host. Since #574 the command nests `shellQuote` three deep (`loginShellWrapped` → `posixShellWrapped` → zmx `sessionCommand`). Fish treats `\'` inside single quotes as an escaped quote; nesting guarantees a `\` before a closing `'`, so the parse shifts and fails. Fish echoes the mangled command (the wall of text) and exits 127, which `SSHReconnectLoop` passes through (only 255 retries).

Verified: `sh -n`/`zsh -n` accept the exact built string, `fish --no-execute` fails. The pre-#574 single-level command parses clean in fish, so this worked before.

## Fix

Only the `loginShellWrapped` layer is fish-parsed; inner layers run under `/bin/sh`. Quote it with `'"'"'` instead of `'\''` — no backslashes, nests safely in both shells. Caveat: fish also collapses `\\` inside single quotes at that layer (`zmxHyperlink`'s `\033\\`).

notSido:
Closes #585

## Summary

Remote surfaces failed when the host's login shell is fish: sshd hands the remote command to the login shell, and the POSIX `'\''` quoting idiom breaks under fish, which interprets `\'` and `\\` inside single quotes. Every new tab on a remote worktree with a fish login shell exited immediately (regression from #574).

The fix introduces `SSHCommand.loginShellQuote`, which quotes tokens so they decode to identical bytes under both fish and POSIX shells: apostrophes become `'"'"'` and backslashes become `'"\\"'` (double-quoted spans, which both parsers collapse identically — no backslash ever appears inside a single-quoted span). It is applied at every layer the remote login shell parses: remote command tokens, forwarded arguments, `env` values, worktree paths, and the zmx connect/reconnect wrappers. Layers parsed by a guaranteed `/bin/sh` (local Ghostty surface command, the `posixShellWrapped` script body) intentionally keep POSIX quoting.

Blocking scripts additionally move to `loginShellWrappedPosixScript`: the login shell now parses only one fish/POSIX-portable line and then `exec`s `/bin/sh` to run the POSIX runner script (`set -u`, `trap`, `$1`), which fish could never parse. The user's script itself still runs under their own login shell inside the runner.

Notes for review:

- **Scope disclosure**: two commits go beyond #585's literal symptom (new remote tabs dying), though both share its root cause — a POSIX-quoted payload handed to a fish login shell. (1) "Run remote blocking scripts under sh" is a behavior change: blocking scripts weren't the reported symptom, but without it run-scripts stay broken on fish hosts. (2) The final commit hardens #756's terminal-compatibility wrap, which landed after this issue was filed and reintroduced the same class of bug in a new layer. Excluding either would make "remote surfaces work when the login shell is fish" false, so they ride along here rather than as separate PRs — happy to split (1) out if preferred.
- Rebased onto main after #756 (remote SSH terminal compatibility). The two fixes are complementary, but #756's new `/bin/sh` wrap quotes its payload with the POSIX-only `shellQuote` while the remote login shell is what parses that string; the final commit switches the wrap to `loginShellQuote` and adds a fish regression test for the full interactive wrap (fish otherwise collapses `\\` inside single quotes and unbalances on a trailing backslash).
- The quoting contract is deliberately narrowed from "fish or csh" to "fish or a POSIX-family shell": the `'"\\"'` backslash spans decode differently under (t)csh. csh support was never tested and is not claimed anymore.
- Contributors need `fish` installed to run `make test` locally (`brew install fish`); the fish regression tests hard-fail rather than silently skip when it is missing, and CI now installs fish in the test jobs. The new install step was exercised on a fork CI run (fish 4.8.1 poured cleanly on `macos-26`); the rest of that run stopped at Tuist auth, which only works in this repo's context.

## Type of change

- [ ] Documentation
- [ ] Other (please describe)

## How was this tested?

New regression tests execute a real `fish` binary (`--no-config`, isolated `HOME`/`XDG_CONFIG_HOME`) and assert byte-for-byte round-trips of adversarial tokens (`argument:it's\\literal`, UTF-8, embedded quotes) through the actual wrapper pipeline: `remoteCommand` under a fish login shell, positional arguments and `env` values, the nested `posixShellWrapped` script, the blocking-script runner (including re-parsing the ssh command line under `/bin/sh` to mirror ssh's argv hand-off), and the worktree `cd` path. CI installs fish so these run on every push. Also verified manually against a remote host with a fish login shell.

- [x] `make test` passes

- Model(s): Claude Opus 4.8, Claude Fable 5
- Harness / tools: Claude Code (development, testing, and pre-PR review)

## Checklist

### DIFF
--- a/.github/actions/setup-macos/action.yml
+++ b/.github/actions/setup-macos/action.yml
@@ -33,7 +33,7 @@ runs:
     - name: Install Fish
       if: ${{ inputs.install-fish == 'true' }}
       shell: bash
-      run: brew install fish
+      run: brew list fish >/dev/null 2>&1 || brew install fish
     - name: Authenticate with Tuist
       if: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository }}
       shell: bash
--- a/.github/actions/setup-macos/action.yml
+++ b/.github/actions/setup-macos/action.yml
@@ -5,6 +5,9 @@ inputs:
   disable-tools:
     description: Comma-separated mise tools to skip, so CI installs only what it needs.
     default: spm:swiftlang/swift-format,npm:create-dmg
+  install-fish:
+    description: Install Fish for shell-compatibility regression tests.
+    default: "false"
 
 runs:
   using: composite
@@ -27,6 +30,10 @@ runs:
       with:
         version: 2026.3.0
         cache: true
+    - name: Install Fish
+      if: ${{ inputs.install-fish == 'true' }}
+      shell: bash
+      run: brew install fish
     - name: Authenticate with Tuist
       if: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository }}
       shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,8 @@ jobs:
         with:
           submodules: recursive
       - uses: ./.github/actions/setup-macos
+        with:
+          install-fish: true
       - parallel:
           - run: make lint
           - run: make inspect-dependencies
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -47,6 +47,8 @@ jobs:
         with:
           submodules: recursive
       - uses: ./.github/actions/setup-macos
+        with:
+          install-fish: true
       - parallel:
           - run: make lint
           - run: make inspect-dependencies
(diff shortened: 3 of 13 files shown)


sbertix (approved):
Thanks 🙇‍♂️

notSido:
> Thanks 🙇‍♂️

Thank you as well for the quick review 😁 
Cheers!

PR #774, opened 2026-08-05
Add an accessibility text size for the app chrome

### LINKED ISSUE #662: Accessibility: large-text options for the app chrome
### Summary

Add an **Accessibility** section to the Settings window with **1–2 large-text options** that scale the *app chrome* — the SwiftUI UI (sidebar worktree list, tab bar, toolbar labels, Settings panes) — independently of the terminal surface.

This is chrome-only. The Ghostty terminal surface already has its own font-size zoom, so this request explicitly does **not** touch terminal rendering.

### Motivation

The app chrome is currently sized by the system default only. Users on high-resolution displays, or with low vision, have no in-app way to enlarge the sidebar/tab/toolbar text without changing the OS-wide display scale (which affects everything, including the terminal). AGENTS.md already mandates *"Use Dynamic Type, avoid hardcoded font sizes"*, so the chrome is well-positioned to respond to a Dynamic Type override.

### Proposed options (1–2, keep it minimal)

1. **Chrome Text Size** — a segmented / stepper control with a small set of discrete choices, e.g. `Default`, `Large`, `Extra Large`, mapped to `DynamicTypeSize` values (`.large` → `.xLarge` → `.xxLarge`) applied to the chrome view hierarchy.
2. *(optional second)* **Bold Chrome Text** — a toggle that applies `.fontWeight(.semibold)`/legibility weight to chrome labels for users who need higher contrast/weight rather than raw size.

Both persist as global settings and take effect live.

### Implementation sketch

- Add an `Accessibility` case to the Settings section list and a matching `AccessibilitySettingsView` (mirror the existing `AppearanceSettingsView` pattern under `SupacodeSettingsFeature/Views/`).
- Add fields to `GlobalSettings` (`SupacodeSettingsShared/Models/GlobalSettings.swift`) — e.g. `chromeTextSize: ChromeTextSize` (new enum in `SupacodeSettingsShared/Models`) and optional `boldChromeText: Bool` — with `decodeIfPresent` defaults so existing settings files migrate cleanly.
- Surface the corresponding state in `SettingsFeature.State` and bind it in the new view.
- Apply the setting to the chrome root by injecting `.dynamicTypeSize(...)` (and, if adopted, a legibility/weight modifier) on the main window's chrome hierarchy — scoped so it does **not** propagate into the Ghostty surface view.

### Acceptance criteria

- [ ] A new **Accessibility** section appears in Settings.
- [ ] Changing **Chrome Text Size** live-resizes sidebar, tab bar, toolbar, and Settings text — but leaves the terminal surface font unchanged.
- [ ] The choice persists across app restarts and older settings files decode without error.
- [ ] No hardcoded font sizes introduced; changes flow through Dynamic Type (per AGENTS.md).

### Non-goals

- Changing the terminal surface font size (already handled by Ghostty zoom).
- A free-form numeric point-size field — keep it to a small set of discrete, system-mapped options.

sageframe-no-kaji:
Closes #662

## Summary

Adds a chrome text size setting so the sidebar, tab bar, toolbars, and Settings can be enlarged without changing the OS-wide display scale, which resizes the terminal along with everything else.

macOS SwiftUI does not resize text in response to `dynamicTypeSize`, so the size is applied by resolving each semantic font explicitly: a `uiTextScale` environment value is published at every window root, and chrome text opts in through `appFont(_:weight:monospaced:)` in place of `.font(.body)`. The Ghostty terminal surface is NSView-backed with its own font system and is unaffected.

Three points that may be worth a closer look during review:

- **The toolbar title publishes the size a second time** inside its `TerminalSchemeHost` closure. That host re-hosts its content in a fresh `NSHostingView`, which starts a new environment rather than inheriting the window's. Views behind the other imperatively-constructed hosting roots (command palette panel, terminal split tree) are deliberately left alone rather than plumbed, to keep this change out of the terminal internals.
- **Section headers need `appFontInheriting`.** `List` styles its own headers and SwiftUI does not expose the font it resolved, so there is nothing to scale in place. The modifier leaves them alone at Default and only supplies a point size above that, so the repository name and the Active and Pinned titles grow with the rows beneath them.
- **`chromeTextSize` is mirrored into `SettingsFeature.State`** like the other global fields, so the picker binds through the store and persists on the existing `.binding` path. Scaled point sizes are rounded to whole points, and the system text-style metrics are read once rather than on every body evaluation.

A missing or unrecognized `chromeTextSize` falls back to the default rather than throwing, which would reset the rest of the settings file — matching the existing `appVisibility` decode. The new pane is reachable at `supacode://settings/accessibility` and `supacode settings accessibility`, like every other pane.

## Type of change

- [ ] Documentation
- [ ] Other (please describe)

## How was this tested?

`make check` and `make test` both pass locally. New tests cover the case order and raw-value stability, the decode migration from a settings file predating the key, the fallback for an unrecognized value, and that an unrelated settings change no longer resets the size.

Also verified by running a Debug build against an isolated data directory at Extra Large: the toolbar title, the sidebar rows, and the Active/Pinned section headers all scale, and the terminal surface is unaffected.

- [x] `make test` passes

## Checklist


sbertix:
Thank you 🙇‍♂️ let me fold in some changes with the new tab bars, add some missing cases, and report back 💪

sbertix (approved):
Thanks 🙇‍♂️

Issue #498, opened 2026-06-26
supacode repo open <path> seems to do nothing


ggascoigne:
I'm having trouble getting supacode cli to do what I think it should, and I suspect that the issue is it doesn't really do what the commands sort of describe.

Specifically I want a command in vscode that will switch to supacode with the active vscode project as the active supacode folder.  Basically mimicking the open in editor command in supecode.

I thought that `supacode repo open <path> would do it, but that appears to be a no-op, I thought that `supacode open` would ensure that supacode is the active osx window, but it also appears to do nothing. Actually that's not quite true, if supacode isn't running, it runs it.

All of this makes me suspect that I've completely misunderstood what the cli does and how it works.  I.e. it might be my fault, it might be a docs bug, or it might be a real issue.

sbertix:
`supacode repo open` is used to add new repositories/folders to Supacode. 

To have Supacode in the foreground you can use `supacode open`, and for selecting a specific worktree/tab/surface you can use `supacode worktree focus -w <worktree_id>`/`supacode tab focus -t <tab_id> -w <worktree_id>`/`supacode surface focus -s <surface_id> -t <tab_id> -w <wortkree_id>`.

ggascoigne:
I'd say that `supacode open` only works some of the time. It does seem to raise the window, but it doesn't give it focus. And if your editor is obscuring the supacode window, then it's still obscuring the window after this command.

As for all of the commands that use ids, how are you meant to find those ids? from looking at the references in issues and code, they are supacode deeplinks and appear to be completely opaque.

sbertix:
There's listing commands for all. I guess installing the skills from `Settings > Deveoper` and having the agents run you through the options etc is probably the easiest way to start.

ggascoigne:
Sweet, I was only using the cli help and it doesn't list the `list` command.  `supacode worktree list` works just fine, thank you.

That said, the focus command does change what's displayed in supacode, but it doesn't seem to actually switch to the app and focus it.

Issue #290, opened 2026-05-05
Feature request: Create worktree from existing branch


isaacharrisholt:
Fairly simple, would be nice not to have to create a new branch when creating a new worktree. I often work on stuff in branches in my default worktree, then want to hand off to agents running in Supacode. There's no easy way to do this, currently!

brodie-euc:
This would be awesome 👍

sbertix:
Hey 👋

I agree that this is probably more urgent than the ability to create worktrees from arbitrary local or remote branches (which was discussed a while back). However, it comes with an additional set of concerns.

Specifically, I’d like to limit this to this case:

> I often work on stuff in branches in my default worktree, then want to hand off to agents running in Supacode.

So, in short, the following:
-  Only the primary worktree. Linked worktrees are already worktrees, so there’s nothing to convert. The trigger for this action is: “I started ad-hoc work in my main checkout and want to hand it off.”
-  Only a named branch (no detached `HEAD`).
-  The branch must differ from the repo default. If you’re already on `main`, there’s nothing meaningful to extract.
-  After handoff, the primary worktree moves back to the default branch. Git won’t check out the same branch in two worktrees, so we need to move the primary off the branch before creating the new worktree for it. Switching to the default branch (instead of detaching) keeps the primary in a clean, named state and matches the “hand off and go back to main” intent.

Wdyt?

isaacharrisholt:
That looks sensible. Honestly, I'd be happy having to create the branch myself, but this seems neater. For context, my current workflow is this:

- Do work in primary checkout (often against `main`)
- Create branch
- `git checkout main`
- Create Supacode worktree
- In worktree, `git checkout <branch I just created>`
- `git branch -D <supacode-created branch>`

It's not the end of the world, it's a small annoyance. Being able to create from arbitrary local or remote branches (without creating a new branch -- could be optional?) would cut out the last two steps, which I would be happy with.

josefmichalek-etnetera:
+1 on this

sbertix:
> Being able to create from arbitrary local or remote branches (without creating a new branch -- could be optional?)

Creating a new branch is super easy. It’s the “without creating a new branch” part that scares me. Do we still run the setup script? Do we copy files according to “Copy ignored/untracked files” plus any existing changes? What happens if the branch is already checked out in a different worktree and we can’t check it out? What state do we leave the previous worktree in? Do we just fail?

[...]

PR #797, opened 2026-08-13
Allow worktree creation from an existing branch; fix repo open's silent success

alexandersmith-294:
Related to #290, #498 — see Summary for why neither uses the closing keyword.

## Summary

Six commits, two related pieces of work:

**Existing-branch reuse.** Worktree creation used to reject any branch name
that already existed, so a branch created outside Supacode — or left behind
after its worktree was deleted — could never be opened again except by
renaming it. `GitClient.branchAvailability` now classifies a branch name as
`absent` / `reusable` / `checkedOut` via one `for-each-ref` call, and
worktree creation offers an explicit "Reuse existing branch" opt-in
(`--reuse-existing-branch` on the CLI) when the branch is unused. A branch
checked out in another worktree is still refused — git only allows one
checkout per branch, and `wt sw` would otherwise silently hand back the
*other* worktree's path.

This PR's own commit on top of that feature fixes a bug found during
verification: `ShellClient` trims trailing whitespace off command output,
so a branch-with-no-worktree line from `for-each-ref` loses its trailing
tab whenever it lands last, and the parser required that tab to find the
name separator — misreporting the last unused branch in the repo as
`absent` and sending reuse into `git worktree add -b` and a "branch already
exists" failure for exactly the branch it was supposed to adopt. Also
routes the resolved reuse directory through `previewWorktreeDirectory`
(always concrete) instead of `resolvedWorktreeDirectory` (nil unless
overridden), since `wt sw` refuses to adopt a branch without a worktree
unless given an explicit `--path`.

**`repo open` contract fixes.** `supacode repo open <path>` answered the
CLI socket `{"ok": true}` the moment the action was dispatched, before
resolution, persistence, or reload had run — so it exited 0 for a path that
resolved to nothing, and for an existing worktree that was never actually
surfaced in the UI. It now defers the ack until the load reports what each
requested path became: adopts + selects a worktree, opens a repository
root, or answers `ok: false` with a reason (non-zero exit). A follow-up
commit fixes a cancellation race the deferred-ack design introduced: every
repository load shares one cancel ID, so a periodic refresh (or a second
`repo open` moments later) could cancel the original load and leave the CLI
call hanging for the full 180s timeout even though the open had actually
succeeded — the requested paths are now parked in state and drained by
whichever load finishes next, not tied to a specific effect surviving
cancellation. `--help` text was updated to describe the new contract.

**Why no closing keyword:** #290 ("create worktree from existing branch")
was explicitly closed `NOT_PLANNED`, and #498 ("`repo open` seems to do
nothing") was closed `COMPLETED` before this branch's fix — despite
`beafa43d`'s commit describing the identical silent-no-op symptom as still
present. Both closures were deliberate triage calls at the time; this PR
references them for context rather than asserting a reversal that isn't
mine to make. Maintainer judgment call on whether either should be
reopened/relabeled.

## Type of change

- [x] Bug fix (`repo open` silent success, cancellation race, branch-availability parsing)
- [x] Feature (existing-branch reuse)
- [ ] Documentation
- [ ] Other (please describe)

## How was this tested?

- `make build-app` — succeeds clean (one pre-existing, unrelated warning in `AppFeature.swift:731`, not touched by this branch).
- Focused run across the tests for both areas above (`GitClientBranchAvailabilityTests`, `RepositoriesFeatureBranchReuseTests`, `RepositoriesFeatureOpenAdoptionTests`, `AppFeatureCommandAckTests`, `RepositoriesFeatureTests`, `WorktreeCreationPromptFeatureTests`, `WorktreeCreationPromptParentTests`, `DeeplinkClientTests`): **516/516 passed**, 0 failures (6 are intentional negative-path tests).
- Full `make test` (all 4 bundles, ~3,145 tests): **3,143 passed, 2 failed.** The 2 failures (`GhosttyRuntimeBundledOverridesTests.backgroundColorTracksColorScheme`, `.initSeedsResolvedColorSchemeBeforeFirstRead`) are in the unrelated terminal-theming suite — confirmed zero diff between this branch and `main` for that test file or the terminal client code, so this is a pre-existing, environment-dependent failure (bundled theme resources resolving light/dark to the same color on the build machine), not a regression from this work.
- `make check` (swift-format + swiftlint) passes clean, no changes needed.

- [ ] `make test` passes — passes except the 2 pre-existing, unrelated Ghostty theming failures noted above
- [ ] I built and ran the app to confirm the change works — verified via the test suites above; did not manually exercise the reuse/open flows in a running app

- Model(s): Claude Sonnet 5, Claude Opus 5
- Harness / tools: Claude Code

## Checklist

- [ ] This pull request is linked to an issue with `Closes #` above.


sbertix:
> repo open contract fixes

This makes total sense. We’re migrating to ack-based completions everywhere, so threading it in here makes sense too. So #498 being marked as “Done” is absolutely non-blocking (as a note, it was done because the issue seemed to conflate `repo open` with the regular `open` command, not because of the issue’s merits).

> Existing-branch reuse

I still have concerns about this as originally proposed in #290. However, I have to admit that this “reusable” split addresses most of them, so I’m happy to review it in action 💪 I’ve reopened the issue and marked it as ready.

Could you reference it with “Closes” in the PR body and reword your commits so they don’t mention any harness, in accordance with the guidelines? Thanks 🙇‍♂️

